Issue #26, opened 2021-01-12
Use assessments rather then 'shared assessments' for display in session builder


conundrumgirl:
/v1/sharedassessments/{guid}/import -  Import an assessment revision into the local app context.

When an assessment being added to the session we need to use endpoint above to first import it into local assessments, prior to adding it to the session
https://developer.sagebridge.org/swagger-ui/index.html#/ has the description, usage and payload

Then the study session should reference the assessment that is not shared.

PR #54, opened 2021-02-06
Import Assessments into Local Context

leonli33:
Before an assessment is used in a session, import it into local assessments. Study session will reference assessments that are not shared.

Resolves: #26.


src/components/studies/session-creator/SessionCreator.tsx
```diff
@@ -262,24 +306,28 @@ const SessionCreator: FunctionComponent<SessionCreatorProps> = ({
...
-              {!getActiveSession(sessions)
-                ? 'Please select group and session'
-                : `Add  to  ${getActiveSession(sessions)?.name} `}
-            </Button>
-          </DialogActions>
+              <Button
+                variant="contained"
+                onClick={async () => {
+                  setIsAddingAssessmentToSession(true)
+                  await updateAssessments(getActiveSession(sessions)!.id, [
```
conundrumgirl, line 317: I think it makes sense to rework this fn call. You are basically creating an array that contains the previous and new assessments and then in your function you are breaking this array up depending on if the assessment is 'previous'.
Why not have a function that takes previous and new assessments. Then you will just need to import the new assessments and then join those 2 arrays

src/components/studies/session-creator/SessionCreator.tsx
```diff
@@ -136,8 +153,30 @@ const SessionCreator: FunctionComponent<SessionCreatorProps> = ({
     })
   }
 
-  const updateAssessments = (sessionId: string, assessments: Assessment[]) => {
+  const updateAssessments = async (
```
conundrumgirl, line 156: so here, just have a (sessionId, prevAssessments, newAssessments) as params.
Then you just need to import/get resources for new assessments and then you can concat those arrays and update the assessments for session

conundrumgirl (changes requested):
Looks good. One proposed change

conundrumgirl (approved):
Nice!

conundrumgirl (approved):

PR #83, opened 2021-03-04
Create Tests for Pagination

leonli33:
Created unit tests that tests the pagination functionality of the participant manager grid.


src/types/types.ts
```diff
@@ -141,6 +141,9 @@ export type ParticipantAccountSummary = {
   externalIds: StringDictionary<string>
   studyExternalId?: string
   status: 'unverified' | 'pending' | 'verified'
+  externalId?: string
```
conundrumgirl, line 144: I don't think we need this here. This should map to what we get from the API

src/components/studies/participants/ParticipantSearch.tsx
```diff
@@ -72,7 +72,7 @@ const useStyles = makeStyles(theme => ({
 }))
```
conundrumgirl, line 72: Did you make any changes in this file or is it just formatting?

conundrumgirl (changes requested):
Nice work on the tests!  One question and one possible correction.

src/components/studies/participants/ParticipantSearch.tsx
leonli33: Just formatting

src/types/types.ts
leonli33: Gotcha. It appears that some results from the API do have an externalID field while others do not:
Screen Shot 2021-03-04 at 9 24 18 AM
v.s.
Screen Shot 2021-03-04 at 9 24 40 AM

I will remove it.

conundrumgirl (approved):
look good

PR #95, opened 2021-03-13
Search for participant on enter key press.

### LINKED ISSUE #94: Participant search should trigger on enter
you should be able to hit 'enter' and start searching.

leonli33:
Search for a participant using ID when the enter key is pressed. Added tests to the search participant functionality.

Resolves #94


conundrumgirl (approved):
looks good

PR #107, opened 2021-03-26
App design

conundrumgirl:
App design

### DIFF
--- a/src/components/studies/StudyList.tsx
+++ b/src/components/studies/StudyList.tsx
@@ -19,7 +19,6 @@ import ConfirmationDialog from '../widgets/ConfirmationDialog'
 import { MTBHeading } from '../widgets/Headings'
 import Loader from '../widgets/Loader'
 import StudyCard from './StudyCard'
-import _ from 'lodash'
 
 type StudyListOwnProps = {}
 
@@ -205,12 +204,7 @@ const StudyList: FunctionComponent<StudyListProps> = () => {
     data: [],
   })
 
-  const { data: studies, status, error, run, setData: setStudies } = useAsync<
-    Study[]
-  >({
-    status: 'PENDING',
-    data: [],
-  })
+
 
   const resetStatusFilters = () =>
     setStatusFilters(sections.map(section => section.status))


src/components/studies/StudyBuilder.tsx
```diff
@@ -293,9 +284,19 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
...
+                        currentAppDesign={
+                          builderInfo.study.clientData.appDesign ||
+                          ({} as StudyAppDesign)
+                        }
+                        onSave={() => {
+                          saveStudy(builderInfo.study)
+                        }}
+                        onUpdate={(data: StudyAppDesign) => {
+                          setHasObjectChanged(true)
+                          builderInfo.study.clientData.appDesign = data
```
conundrumgirl, line 296: you don't want to do it like this. because you re modifying your builderInfo twice: once lne 296 and then in setData. Something like this would work better:
```  
onUpdate={(data: StudyAppDesign) => {
                          setHasObjectChanged(true)
                          const updatedStudy = {...builderInfo.study}
                          updatedStudy.clientData.appDesign = data
              
                          setData({
                            ...builderInfo, study:  updatedStudy
                          })
         ```

File: src/components/studies/app-design/ContactInformation.tsx [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+    alignItems: 'center',
+  },
+  row: {
+    flexDirection: 'row',
+    display: 'flex',
+    alignItems: 'center',
+    width: '80%',
+  },
+  icon: {
+    marginRight: '10px',
```
conundrumgirl, line 20: we are doing in multiples of 8 if possible here and in others theme.spacing(n)

src/components/studies/app-design/StudySummaryRoles.tsx
```diff
@@ -0,0 +1,20 @@
...
+  type: String
+  name: String
+}
+
+const StudySummaryRoles: React.FunctionComponent<StudySummaryRoleProps> = ({
+  type,
+  name,
+}) => {
+  return (
+    <div style={{ marginTop: '20px', wordWrap: 'break-word' }}>
```
conundrumgirl, line 14: can it be 16 or 24?

PR #118, opened 2021-04-09
Made dropdown from Material UI components 

leonli33:
Screen Shot 2021-04-09 at 10 59 35 AM

### DIFF
--- a/src/components/studies/app-design/LeadInvestigatorDropdown.tsx
+++ b/src/components/studies/app-design/LeadInvestigatorDropdown.tsx
@@ -117,7 +117,7 @@ const LeadInvestigatorDropdown: React.FunctionComponent<LeadInvestigatorDropdown
       <Select
         labelId="lead-investigator-drop-down"
         id="lead-investigator-drop-down"
-        value={currentInvestigatorSelected || 'Select pricinple investigator'}
+        value={currentInvestigatorSelected}
         onChange={e => {
           onChange(e.target.value)
         }}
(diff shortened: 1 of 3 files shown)


src/components/studies/app-design/LeadInvestigatorDropdown.tsx
```diff
@@ -107,43 +109,54 @@ const LeadInvestigatorDropdown: React.FunctionComponent<LeadInvestigatorDropdown
...
+            horizontal: 'center',
+          },
+          transformOrigin: {
+            vertical: 'top',
+            horizontal: 'center',
+          },
+        }}
+      >
+        <MenuItem
+          value="Select pricinple investigator"
```
conundrumgirl, line 144: spelling? Also, what happens when nothing is selected? We want to send '"Select pricinple investigator" to the parent control.

src/components/studies/app-design/LeadInvestigatorDropdown.tsx
```diff
@@ -107,43 +109,54 @@ const LeadInvestigatorDropdown: React.FunctionComponent<LeadInvestigatorDropdown
...
+          onChange(e.target.value)
+        }}
+        className={classes.container}
+        disableUnderline
+        classes={{
+          selectMenu: classes.selectMenu,
+          root: classes.selectPrincipleInvestigatorButton,
+        }}
+        MenuProps={{
+          classes: { list: classes.listPadding, paper: classes.listBorder },
```
conundrumgirl, line 131: I don't know if you can, but if it is possible to just have a style property here rather than class, since it is only a single prop (border, padding) it would be ideal. I

conundrumgirl (changes requested):

src/components/studies/app-design/LeadInvestigatorDropdown.tsx
leonli33: Ah, i assumed if nothing was selected we wanted to send an empty string to the parent control. I will fix that.

conundrumgirl (approved):

PR #119, opened 2021-04-11
Added blue border as hover state to study card

leonli33:
Added a blue border to study cards when the user hovers over them. Also, set the blue border to appear on newly created study cards for only 600ms.

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -70,6 +70,7 @@ export type AssessmentResource = {
   url: string
   version: number
 }
+
 export type Assessment = {
   createdOn: string
   customizationFields: object
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -69,8 +69,8 @@ export type AssessmentResource = {
   upToDate: boolean
   url: string
   version: number
-}
 
+}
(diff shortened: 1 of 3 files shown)


src/types/types.ts
```diff
@@ -213,6 +210,11 @@ export type Phone = {
   nationalFormat?: string
 }
 
+export type HighlightStudyCardWithBlueBorder = {
```
conundrumgirl, line 213: types.ts are for shared types that affect business logic. The types used for UI can/should live in the relevant files.

conundrumgirl (changes requested):
Is there a reason why we are not doing it via CSS using :hover?

leonli33:
> Is there a reason why we are not doing it via CSS using :hover?

Yea. So when the user hovers over a certain card, I want to execute javascript. More specifically, I need to clear the setTimeout to ensure that not more than two studies are highlighted at once.

src/types/types.ts
leonli33: Alright sounds good. I needed to use the same type in more than one file so I thought that it might be best to put it here.

PR #124, opened 2021-04-16
Restyled Study Without Scheduling Page 

### LINKED ISSUE #121: Study without schedule -- styling
When study is just created and there is no schedule  there is an intro screen that needs some CSS attention.
Clicking on 'NoSchedule Study' will take you to the right page

image
need to look like this:
image

leonli33:
Page now looks like this:
Screen Shot 2021-04-16 at 4 50 53 PM

Screen Shot 2021-04-16 at 4 51 25 PM

Resolves #121

### DIFF
--- a/src/components/studies/scheduler/Scheduler.tsx
+++ b/src/components/studies/scheduler/Scheduler.tsx
@@ -4,7 +4,7 @@ import {
   createStyles,
   FormControlLabel,
   makeStyles,
-  Theme
+  Theme,
 } from '@material-ui/core'
 import SaveIcon from '@material-ui/icons/Save'
 import _ from 'lodash'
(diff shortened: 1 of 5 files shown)


src/components/studies/scheduler/Scheduler.tsx
```diff
@@ -56,6 +56,13 @@ const useStyles = makeStyles((theme: Theme) =>
       justifyContent: 'space-between',
       paddingRight: theme.spacing(2),
     },
+    studyStartDateContainer: {
```
leonli33, line 59: There are a lot of formatting changes in this file, but this is the only thing that I actually added.

src/components/studies/scheduler/Scheduler.tsx
```diff
@@ -144,40 +153,43 @@ const Scheduler: FunctionComponent<
...
+              onClick={() => onSave()}
+              startIcon={<SaveIcon />}
+            >
+              Save changes
+            </Button>
+          )}
+        </div>
+        <Box bgcolor="#fff" p={2} mt={3} key="scheduler">
+          <TimelinePlot something=""></TimelinePlot>
+          <div className={classes.studyStartDateContainer}>
```
leonli33, line 188: The style added above is used here.

PR #131, opened 2021-04-20
Fix study duration screen display based on screen size

leonli33:
The screen now looks like:

When the height is shortened:
Screen Shot 2021-04-19 at 10 28 35 PM

When the width is shortened:
Screen Shot 2021-04-19 at 10 30 54 PM

On Mobile:
Screen Shot 2021-04-19 at 10 31 48 PM

### DIFF
--- a/src/components/studies/StudyBuilder.tsx
+++ b/src/components/studies/StudyBuilder.tsx
@@ -83,6 +83,10 @@ const useStyles = makeStyles((theme: ThemeType) => ({
     alignItems: 'center',
     backgroundColor: '#FAFAFA',
     height: '90vh',
+    minHeight: '700px',
+    [theme.breakpoints.down('md')]: {
+      height: '100vh',
+    },
   },
 }))
 
--- a/src/components/studies/StudyBuilder.tsx
+++ b/src/components/studies/StudyBuilder.tsx
@@ -80,7 +80,7 @@ const useStyles = makeStyles((theme: ThemeType) => ({
     textAlign: 'center',
     backgroundColor: '#FAFAFA',
     height: '100vh',
-    paddingTop: '12vh',
+    paddingTop: theme.spacing(18.5),
   },
 }))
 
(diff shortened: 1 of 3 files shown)


src/components/studies/StudyBuilder.tsx
```diff
@@ -83,6 +83,10 @@ const useStyles = makeStyles((theme: ThemeType) => ({
     alignItems: 'center',
     backgroundColor: '#FAFAFA',
     height: '90vh',
```
conundrumgirl, line 85: why do you need to specify the height?

leonli33: Without the height the screen looks like this: 
Screen Shot 2021-04-20 at 2 28 47 PM

src/components/studies/StudyBuilder.tsx
```diff
@@ -77,12 +77,10 @@ const useStyles = makeStyles((theme: ThemeType) => ({
...
   introInfoContainer: {
-    display: 'flex',
-    flexDirection: 'row',
-    justifyContent: 'center',
-    alignItems: 'center',
+    textAlign: 'center',
     backgroundColor: '#FAFAFA',
-    height: '90vh',
+    height: '100vh',
+    paddingTop: '12vh',
```
conundrumgirl, line 83: I don't think we need to have the vh units for top padding. Since the height of the element is not in vh. let's just stick to whatever margin is in the design in multiples of 8

src/components/studies/scheduler/IntroInfo.tsx
```diff
@@ -22,12 +23,12 @@ const useStyles = makeStyles((theme: Theme) =>
       fontSize: '18px',
       fontStyle: 'normal',
```
conundrumgirl, line 24: we only need to specify fontStyle when it's not 'normal'. I don't think we should use % on the margins/padding/etc
and even in width for most cases

conundrumgirl (changes requested):
a couple of small tweaks

conundrumgirl (approved):

PR #140, opened 2021-04-28
Redo app design page to comply with updated study object.

leonli33:


src/components/studies/StudyBuilder.tsx
```diff
@@ -162,14 +157,17 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
...
+
     try {
       const newVersion = await StudyService.updateStudy(study, token!)
       const updatedStudy = { ...study, version: newVersion }
       setData({
         ...builderInfo,
         study: updatedStudy,
       })
-
+      if (error !== '') {
```
conundrumgirl, line 168: do we need this check?

src/components/studies/StudyBuilder.tsx
```diff
@@ -391,24 +390,19 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
...
                         saveLoader={saveLoader}
                         id={id}
-                        currentAppDesign={
-                          builderInfo.study.clientData?.appDesign ||
-                          ({} as StudyAppDesign)
-                        }
-                        onSave={() => {
+                        study={builderInfo.study}
+                        onSave={(updatedStudy: Study) => {
+                          setData({
```
conundrumgirl, line 395: saveStudy does setData. Do we need to call setData again?

conundrumgirl (approved):

src/components/studies/StudyBuilder.tsx
leonli33: Yes, if we do not do this then the error message will still show on top of the screen even if the study was updated successfully.

src/components/studies/StudyBuilder.tsx
leonli33: Updated.

src/components/studies/StudyBuilder.tsx
conundrumgirl: What I'm questioning is the needs to check.  I think just doing setError('') is sufficient. Yes it's extra set if it's already empty but it removes the check an decreases mental load for a person who might need to look at the code later. It's really a tiny detail.

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -610,9 +610,9 @@ const AppDesign: React.FunctionComponent<
       const formattedPhone = formatPhoneNumber(phoneWithoutZipcode)
       setIrbPhoneNumber(formattedPhone)
     }
-    const isWelcomeScreenDataIsEmpty =
+    const isWelcomeScreenDataEmpty =
       Object.keys(study.clientData.welcomeScreenData || {}).length == 0
-    const welcomeScreenData = isWelcomeScreenDataIsEmpty
+    const welcomeScreenData = isWelcomeScreenDataEmpty
```
conundrumgirl, line 615: can it just be `welcomeScreenData  =  Object.keys(study.clientData.welcomeScreenData || {}).length == 0`

PR #141, opened 2021-05-04
Update drawer menu

### LINKED ISSUE #134: Styling of the mobile 'hamburger' menu
We need to update the hamburger menu for studies as well as the main menu

image

image
Also tagged you in figma

leonli33:
Side menu now looks like this:
When the user is not signed in:
Screen Shot 2021-05-04 at 4 20 35 PM
When the user is not viewing a study:
Screen Shot 2021-05-04 at 4 21 24 PM
When the user is viewing a study:
Screen Shot 2021-05-04 at 4 21 48 PM
When the user hovers over the mtb logo:
Screen Shot 2021-05-04 at 4 22 09 PM

Resolves #134


src/components/widgets/MobileDrawerMenuHeader.tsx
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const MobileDrawerMenuHeader: React.FunctionComponent<MobileDrawHeaderProps> = ({
+  setIsMobileOpen,
+  type,
+}) => {
+  const classes = userStyles()
+  const [isLogoHovered, setIsLogoHovered] = React.useState(false)
+  const logo = (
+    <img
```
conundrumgirl, line 50: since MTBLogo is and svg you should be able to change the color via css :hover.
Alternatively if that doesn't work you can have div instead of an image and have backgrounds differ on :hover in css.  I think it would be preferable to avoid setting state on hover in javascript and instead do it in css using :hover

conundrumgirl (changes requested):

conundrumgirl (approved):

PR #172, opened 2021-06-04
Added notification errors

### LINKED ISSUE #168: Schedule validation for notifications
There are instances when errors are associated with the notifications

image
image

Error messages that include notifications should show up in notifications

leonli33:
Looks like:
Screen Shot 2021-06-04 at 12 00 52 PM

Resolves #168

### DIFF
--- a/src/components/studies/scheduler/SchedulableSingleSessionContainer.tsx
+++ b/src/components/studies/scheduler/SchedulableSingleSessionContainer.tsx
@@ -7,10 +7,9 @@ import { DEFAULT_NOTIFICATION } from '../../../services/study.service'
 import { ThemeType } from '../../../style/theme'
 import {
   AssessmentWindow as AssessmentWindowType,
-
-
-  ScheduleNotification, SessionSchedule,
-  StudySession
+  ScheduleNotification,
+  SessionSchedule,
+  StudySession,
 } from '../../../types/scheduling'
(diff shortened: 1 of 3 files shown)


src/components/studies/scheduler/SchedulableSingleSessionContainer.tsx
```diff
@@ -92,6 +106,21 @@ const SchedulableSingleSessionContainer: FunctionComponent<SchedulableSingleSess
     setWindowErrors(windowErrorsArray)
   }, [sessionErrorState?.sessionWindowErrors])
 
+  React.useEffect(() => {
+    if (!sessionErrorState || sessionErrorState.notificationErrors.size == 0) {
+      setNotificationErrors([])
+      return
+    }
+    const notificationErrorsArray: notificationErrorArrayType[] = []
```
conundrumgirl, line 114: while what you are doing it correct, I think map method is more appropriate because you are transforming one array to the other (sessionErrorState.notificationErrors -> notificationErrorsArray) so I think something like this would be clearer: 
```
const notificationErrorsArray =  sessionErrorState.notificationErrors.map((el, key) => ({  notificationName: 'notification ' + key,  notficationError: el }) )
 ```

conundrumgirl (changes requested):
one nitpicky thing, looks good otherwise

conundrumgirl (approved):

PR #186, opened 2021-06-14
Redid logic for AppDesign page.

leonli33:
All of the text inputs are updating the study object locally instead of doing it all through one method.


src/components/studies/app-design/GeneralContactAndSupportSection.tsx
```diff
@@ -112,9 +106,6 @@ const GeneralContactAndSupportSection: React.FunctionComponent<GeneralContactAnd
                 contactLeadInfo: newContactLeadObject,
               })
             }}
-            onBlur={() =>
-              updateAppDesignInfo(AppDesignUpdateTypes.UPDATE_STUDY_CONTACTS)
-            }
             multiline
```
conundrumgirl, line 109: please remove unnecessary 
 `multiline rows={1} rowsMax={1}`  and make those props optional on the text input wrapper

src/components/studies/app-design/IrbBoardContactSection.tsx
```diff
@@ -278,9 +275,11 @@ const IrbBoardContactSection: React.FunctionComponent<IrbBoardContactSectionProp
...
             }}
-            onBlur={() =>
-              updateAppDesignInfo(AppDesignUpdateTypes.UPDATE_STUDY_IRB_NUMBER)
-            }
+            onBlur={() => {
+              const newStudy = { ...study }
+              newStudy.irbProtocolId = appDesignProperties.irbProtocolId
+              onUpdate(newStudy)
+            }}
             multiline
```
conundrumgirl, line 283: here and below - you don't need multiline, rows, rowsMax

src/components/studies/app-design/WelcomeScreenMessagingSection.tsx
```diff
@@ -43,21 +42,31 @@ const useStyles = makeStyles(theme => ({
 type WelcomeScreenMessagingSectionProps = {
   appDesignProperties: StudyAppDesign
   setAppDesignProperties: Function
-  updateAppDesignInfo: Function
   saveLoader: boolean
   saveInfo: Function
   SimpleTextInputStyles: React.CSSProperties
+  onUpdate: Function
+  study: Study
```
conundrumgirl, line 49: here and in the other controls. I don't think you should pass down the entire study object. You should only pass the things you are actually updating, welcomeScreenHeader, salutation etc.

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -564,19 +511,20 @@ const AppDesign: React.FunctionComponent<
...
-                      const newWelcomeScreenData = {
-                        ...prevState.welcomeScreenInfo,
-                      }
-                      newWelcomeScreenData.isUsingDefaultMessage =
-                        !prevState.welcomeScreenInfo.isUsingDefaultMessage
-                      return {
-                        ...appDesignProperties,
-                        welcomeScreenInfo: newWelcomeScreenData,
-                      }
+                  onChange={() => {
```
conundrumgirl, line 514: on Change should be called with the actual params of the things that you are changing rather than updating the appDesignProperties reference. I think you should get away from using the appDesignProperties. Just updated the straight up values

conundrumgirl (changes requested):

PR #188, opened 2021-06-14
Added function and alert that checks if the schedule is default

leonli33:
If you want to try it out, I added the following code into the `scheduler.tsx` file:

` 
 <button
      onClick={() => {
            const isSame = isSameAsDefaultSchedule(_schedule)
            alert('same as default ' + isSame)
          }}
          style={{ width: '40px', height: '40px', backgroundColor: 'red' }}
      ></button>
`

Also added this as an optional alert to launch alerts.
Screen Shot 2021-06-14 at 2 43 41 PM

### DIFF
--- a/src/components/studies/scheduler/utility.ts
+++ b/src/components/studies/scheduler/utility.ts
@@ -15,20 +15,15 @@ export function getDropdownTimeItems(): StringDictionary<string> {
 }
 
 export function isSameAsDefaultSchedule(schedule: Schedule) {
-  const defaultSessionDelay = undefined
-  const defaultSessionInterval = undefined
-  const defaultSessionOccurences = undefined
-  for (const session of schedule.sessions) {
-    const sessionDelay = session.delay
-    const sessionInterval = session.interval
-    const sessionOccurences = session.occurrences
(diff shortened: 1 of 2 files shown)


src/components/studies/scheduler/utility.ts
```diff
@@ -13,3 +14,30 @@ export function getDropdownTimeItems(): StringDictionary<string> {
   return menuItems
 }
 
+export function isSameAsDefaultSchedule(schedule: Schedule) {
+  const defaultSessionDelay = undefined
```
conundrumgirl, line 18: it's correct but I think it could be much shorter. e.g.
instead of 
`const defaultSessionDelay = undefined ...
 for (const session of schedule.sessions) {    
  ....const sessionDelay = session.delay ...
  ... 
 const isScheduleDifferentFromDefault =   sessionDelay !== defaultSessionDelay ||`
  just 
  ```  for (const session of schedule.sessions) {    
   const isScheduleDifferentFromDefault = session.delay !== undefined ...```
   
   you can put in the comments that undefined is a default value

conundrumgirl (changes requested):
The function is correct but I think we can simplify it

conundrumgirl (approved):

PR #189, opened 2021-06-14
Preview Study Screen Updates.

leonli33:
Now looks like:
Screen Shot 2021-06-14 at 4 06 18 PM

### DIFF
--- a/src/components/studies/StudyBuilder.tsx
+++ b/src/components/studies/StudyBuilder.tsx
@@ -488,7 +488,7 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
                       <Preview
                         studyId={builderInfo.study.identifier}
                         token={token!}
-                        schedule={builderInfo.schedule}
+                        scheduleSessions={builderInfo.schedule.sessions}
                       ></Preview>
                     )}
                     {section === 'launch' && (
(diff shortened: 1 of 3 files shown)


src/components/studies/preview/Preview.tsx
```diff
@@ -74,12 +78,36 @@ const useStyles = makeStyles((theme: ThemeType) => ({
...
+    width: '24px',
+    marginTop: theme.spacing(0.75),
+  },
 }))
 
 export interface PreviewProps {
   children?: ReactNode
   studyId: string
   token: string
+  schedule: Schedule
```
conundrumgirl, line 110: why not just pass sessions rather than schedule

src/components/studies/preview/Preview.tsx
```diff
@@ -118,6 +151,25 @@ const Preview: React.FunctionComponent<PreviewProps> = ({
       handleError(e!)
     }
   }
+
+  useEffect(() => {
+    const sessions = schedule.sessions
+    const allUniqueAssessments: Assessment[] = []
+    for (const session of sessions) {
```
conundrumgirl, line 158: This works. But as a fun exercise - see if you can accomplish this using `reduce`

conundrumgirl (changes requested):
One small request. and one potential suggestion because why not:)

src/components/studies/StudyBuilder.tsx
```diff
@@ -488,6 +488,7 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
                       <Preview
                         studyId={builderInfo.study.identifier}
                         token={token!}
+                        schedule={builderInfo.schedule}
```
conundrumgirl, line 491: you can just pass the sessions here - don't need the full schedule

conundrumgirl (approved):

PR #193, opened 2021-06-16
Disable Nav menu when one session has no assessments

leonli33:
Looks like:
Screen Shot 2021-06-16 at 1 58 50 PM

Also, when there is only one session remaining, revoke ability to delete the session:
Screen Shot 2021-06-16 at 1 59 25 PM

### DIFF
--- a/src/components/studies/StudyBuilder.tsx
+++ b/src/components/studies/StudyBuilder.tsx
@@ -324,6 +324,7 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
       id={id}
       currentSection={section}
       onNavigate={(section: StudySection) => changeSection(section)}
+      disabled={!allSessionsHaveAssessments()}
     ></NavButtons>
   )
   if (builderInfo.study && !builderInfo.schedule) {
--- a/src/components/studies/StudyBuilder.tsx
+++ b/src/components/studies/StudyBuilder.tsx
@@ -268,7 +268,7 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
   }
 
   const changeSection = async (next: StudySection) => {
-    if (section === next) {
+    if (section === next || !allSessionsHaveAssessments()) {
       return
     }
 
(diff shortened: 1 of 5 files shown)


src/components/studies/StudyBuilder.tsx
```diff
@@ -138,6 +139,23 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
   const setData = (builderInfo: StudyInfoData) => {
     studyDataUpdateFn({ type: 'SET_ALL', payload: builderInfo })
   }
+
+  useEffect(() => {
```
conundrumgirl, line 143: This should not be in the state  as you can derive this value from what you already have in the state. Just make this a function that returns a boolean and call this function where you reference that state var.
also something like this seems simpler

```
return sessions && sessions.length && !sessions.find(session=> !session.assessments || session.assessments.length === 0)
```
we can simplify it farther by using lodash method isEmpty since we already have lodash (_) included in this file.
 `return !_.isEmpty(sessions) &&  !sessions.find(session=>_.isEmpty(session.assesments))`

src/components/studies/StudyLeftNav.tsx
```diff
@@ -162,10 +172,23 @@ const StudyLeftNav: FunctionComponent<StudyLeftNavProps> = ({
               >
                 <img
                   src={typeOfIcon(index, sectionLink.path)[index]}
-                  className={classes.navIcon}
+                  className={clsx(
```
conundrumgirl, line 175: what is the code that actually prevents it from being able to navigate to?

conundrumgirl (changes requested):
We also need to disable next button so we can't navigate to scheduler

src/components/studies/StudyLeftNav.tsx
leonli33: line 149 is where all the nav sections are being disabled.

conundrumgirl (approved):

PR #202, opened 2021-06-18
Update IRB Fields on the study object

leonli33:
Persists information regarding the start/end dates:
Screen Shot 2021-06-18 at 1 51 07 PM

### DIFF
--- a/src/components/studies/launch/IrbDetails.tsx
+++ b/src/components/studies/launch/IrbDetails.tsx
@@ -169,7 +169,7 @@ const IrbDetails: React.FunctionComponent<IrbDetailsProps> = ({
       investigator.affiliation &&
       study.irbProtocolId &&
       irb.name &&
-      study.clientData.irbProtocolTitle
+      study.irbName
     onEnableNext(inputFieldsCorrectFormat)
   })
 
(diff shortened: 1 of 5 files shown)


src/components/studies/launch/IrbDetails.tsx
```diff
@@ -192,10 +186,27 @@ const IrbDetails: React.FunctionComponent<IrbDetailsProps> = ({
...
-    studyData?.irbApprovalDate &&
-    studyData?.irbApprovedUntil &&
-    studyData?.irbApprovalDate > studyData?.irbApprovedUntil
+    study.irbDecisionOn &&
+    study.irbExpiresOn &&
+    study.irbDecisionOn > study.irbExpiresOn
+  const irbDecisionIsApproved =
+    !study.irbDecisionType || study.irbDecisionType === 'approved'
+  const irbDecisionDate = study.irbDecisionOn
+    ? getDateWithTimeZone(new Date(study.irbDecisionOn))
```
conundrumgirl, line 205: I think toLocaleDateString() or one of the moment functions should do the trick. I  don't think we need to the custom function here

conundrumgirl (changes requested):
one question about custom date function. Otherwise looks good

src/components/studies/launch/IrbDetails.tsx
conundrumgirl: OK - we might revisit it later -- but I can see the complexity. looks good

Issue #207, opened 2021-06-22
Access Settings - add user update


conundrumgirl:
![image](https://user-images.githubusercontent.com/5190623/122946767-7e8c1c80-d32e-11eb-90f7-13a12c63c48d.png)
When you add a user (researcher) in the access settings the role selections need to be updated to reflect the design (I tagged you) 
When 'MAKE CO-ADMINISTRATOR' toggle is not checked -- replace the 'access settings' radio with text. When the toggle is checked - set it to selected. Don't worry about the other toggles .

PR #210, opened 2021-06-22
Added Access Settings

leonli33:
Page now looks like:
Screen Shot 2021-06-22 at 12 53 53 PM
Screen Shot 2021-06-22 at 12 54 09 PM

Resolves: #207


conundrumgirl (changes requested):
One small change

conundrumgirl (approved):

PR #213, opened 2021-06-23
Display Real Number of Participants

leonli33:
If the study is out of the design phase, then display the number of enrolled participants:
Screen Shot 2021-06-23 at 10 08 33 AM


conundrumgirl (approved):
Nice work!

Issue #208, opened 2021-06-22
Send SMS link functionality


conundrumgirl:
In Participant Manager -- please add the send  SMS link functionality. It is very similar to deleting users. The endpoint is not complete yet.  There are a few additional changes that need to be made to the delete and sms modals

- select all: need a service function that would select external id, phone if they have it for *all* active/withdrawn/test participants
-  external id in a modal display. Apply formatting that strips the study id and only show the second part as we do in other places
- limit the height and  scroll if the list is too long
- only show the SMS for phone enrolled

PR #221, opened 2021-06-25
Implement Send SMS functionality

leonli33:
Added UI for send SMS functionality, still waiting for backend support
Looks like:
Screen Shot 2021-06-25 at 11 22 53 AM

Also fixed icon error:
Screen Shot 2021-06-25 at 11 38 05 AM

Resolves: #208


src/services/participants.service.ts
```diff
@@ -313,6 +313,34 @@ async function getParticipants(
   return { items: resultItems, total: result.total }
 }
 
+async function getAllParticipantsInEnrollmentType(
```
conundrumgirl, line 316: why do you need the pageSize and offset here if you are getting all

src/components/studies/participants/DialogContents.tsx
```diff
@@ -0,0 +1,248 @@
...
+    return (
+      <Box className={classes.root}>
+        <Box textAlign="center" mx="auto">
+          <CircularProgress />
+        </Box>
+      </Box>
+    )
+  }
+
+  const idsToRemoveList = formatIds(
```
conundrumgirl, line 173: is this just for removal or for sms also? If for both -- please rename the var

src/components/studies/participants/DialogContents.tsx
```diff
@@ -0,0 +1,248 @@
...
+    setLoadingData(false)
+  }
+
+  useEffect(() => {
+    setLoadingData(true)
+    if (selectingAll) {
+      setData()
+      return
+    }
+    const formattedData = selectedParticipants.map(participant => {
```
conundrumgirl, line 134: don't you only need Id's and/or phones? What is the purpose of other data?

src/components/studies/participants/DialogContents.tsx
```diff
@@ -0,0 +1,248 @@
...
+        'all',
+        true,
+      )
+      const resultWithdrawn = await ParticipantService.getAllParticipantsInEnrollmentType(
+        study.identifier,
+        token,
+        'withdrawn',
+        false,
+      )
+      const dataAllTestMembers = resultAll.items
```
conundrumgirl, line 107: those are all memebers, not just test, right? please rename if so

src/components/studies/participants/DialogContents.tsx
```diff
@@ -0,0 +1,248 @@
...
+      const dataWithdrawnMembers = resultWithdrawn.items
+      let testParticipants = dataAllTestMembers.filter(
```
conundrumgirl, line 109: you don't need to use external ids. All participants have the id -- just filter on that

conundrumgirl (changes requested):

PR #229, opened 2021-06-29
Added Batch Edit Feature

leonli33:
Added ability to edit more than one participant at a time:
Screen Shot 2021-06-29 at 12 37 33 PM


src/components/studies/participants/BatchEditForm.tsx
```diff
@@ -0,0 +1,95 @@
+import React from 'react'
+import { Dialog } from '@material-ui/core'
+import { EditParticipantForm } from './ParticipantForms'
+import { EditDialogTitle } from './ParticipantTableGrid'
+import ParticipantService from '../../../services/participants.service'
+
+type BatchEditFormProps = {
+  isEnrolledById: boolean
+  isBatchEditOpen: boolean
+  setIsBatchEditOpen: Function
```
conundrumgirl, line 10: since those are callback functions, it's customary to have them start with on e.g. onSetIsBatchEditOpen

src/components/studies/participants/BatchEditForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+          setIsBatchEditOpen(false)
+        }}
+        shouldWithdraw={false}
+        batchEdit
+      />
+      <EditParticipantForm
+        isEnrolledById={isEnrolledById}
+        onCancel={() => {
+          setIsBatchEditOpen(false)
+        }}
```
conundrumgirl, line 48: This is too long of a function to be inlined. Please have it as a separate function that takes an update type as an pargument

src/components/studies/participants/BatchEditForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+          setIsLoading(true)
+          let participants = selectedParticipants
+          if (isAllSelected) {
+            const resultEnrolled = await ParticipantService.getAllParticipantsInEnrollmentType(
+              studyId,
+              token,
+              'enrolled',
+              false,
+            )
+            const enrolledNonTestParticipants = resultEnrolled.items
```
conundrumgirl, line 60: The variable naming on ln54 and 60 are a little misleading. Ln60 should be same as 54 but just Id's right?

src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -401,8 +416,14 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
...
 
   //trigger data refresh on updates
-  const [refreshParticipantsToggle, setRefreshParticipantsToggle] =
-    React.useState(false)
+  const [
+    refreshParticipantsToggle,
+    setRefreshParticipantsToggle,
+  ] = React.useState(false)
+
+  const toggleParticipantRefresh = () => {
```
conundrumgirl, line 424: it's just a one- line function just do it inline for your control

src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -856,7 +897,16 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
             </CollapsibleLayout>
           </Box>
         </Box>
-
+        <BatchEditForm
+          isEnrolledById={false}
```
conundrumgirl, line 901: while it is true that it will be false since this will only show when they are enrolled by phone, it would be more consistent to check using isSignInById(study.signInTypes)

src/services/participants.service.ts
```diff
@@ -1,4 +1,4 @@
-import _ from 'lodash'
+import _, { constant } from 'lodash'
```
conundrumgirl, line 1: do you need this 'constant' import?

src/components/studies/participants/BatchEditForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+            participants = enrolledNonTestParticipants.map(
+              el => el.participant.identifier,
+            )
+          }
+          for (const selectedParticipantId of participants) {
```
conundrumgirl, line 65: you don't need to wait for one call to complete to make the next. So it would make more sense here to use map instead of the 'for' loop to transform the array of participantIds into array of promises and then do a single away for Promise.all
Same with updating clinic visit.
Also the action (if you are updating note or the date) would apply to all of the participants therefore it should be outside of the loop/map

conundrumgirl (changes requested):
Looks good. A few changes

src/components/studies/participants/BatchEditForm.tsx
leonli33: Line 60 is the entire participant object, not just ids. Line 54 is an object that has the participant data but also the total number of participants (which is not needed in our case).

conundrumgirl (approved):

PR #230, opened 2021-06-29
Restyle study card menu

leonli33:
Now looks like: 
Screen Shot 2021-06-29 at 3 40 58 PM

Resolves: https://sagebionetworks.jira.com/browse/MTB-177

### DIFF
--- a/src/components/studies/StudyList.tsx
+++ b/src/components/studies/StudyList.tsx
@@ -5,7 +5,7 @@ import {
   Divider,
   makeStyles,
   Menu,
-  MenuItem
+  MenuItem,
 } from '@material-ui/core'
 import Link from '@material-ui/core/Link'
 import React, { FunctionComponent, useEffect } from 'react'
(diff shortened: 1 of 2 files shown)


src/components/studies/StudyList.tsx
```diff
@@ -184,6 +204,10 @@ const StudySublist: FunctionComponent<StudySublistProps> = ({
               }}
               isNewlyAddedStudy={highlightedStudyId === study.identifier}
               section={item.sectionStatus}
+              isMenuOpen={
+                menuAnchor !== null &&
+                menuAnchor.study.identifier === study.identifier
```
conundrumgirl, line 209: I think menuAnchor?.study?.identifier === study.identifier is sufficient because if it will resolve to undefined it will not be equal (dont need menuAnchor !== null)

conundrumgirl (changes requested):

conundrumgirl (approved):

PR #248, opened 2021-07-07
Update search functionality

leonli33:
Search for participants in a study based on external_id and phone number.
Screen Shot 2021-07-07 at 4 30 19 PM

### DIFF
--- a/src/components/studies/participants/ParticipantManager.tsx
+++ b/src/components/studies/participants/ParticipantManager.tsx
@@ -540,11 +540,10 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
 
     let participants: ParticipantAccountSummary[] = []
     let total = 0
-    const seenIds = new Set<string>()
     for (const result of results) {
       for (const participant of result.items) {
-        if (seenIds.has(participant.id)) continue
-        seenIds.add(participant.id)
+        const isInList = participants.find((el) => el.id === participant.id) !== undefined
+        if (isInList) continue
         total++
         participants.push(participant)
       }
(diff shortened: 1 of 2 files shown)


src/services/participants.service.ts
```diff
@@ -428,6 +428,53 @@ async function getEnrollmentById(
...
+      : 'all'
+  const noneOfGroups = []
+  const allOfGroups = []
+  if (participantType !== 'TEST') {
+    noneOfGroups.push('test_user')
+  } else {
+    allOfGroups.push('test_user')
+  }
+  let body
+  if (searchType === 'EXTERNAL_ID') {
```
conundrumgirl, line 456: you can get rid of some repetition here. The only difference between external_id and phone search is externalId filter or a phone filter.  So you can create the object outside the condition and then 
have an 'if' statement to set the relevant field

src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -516,29 +520,62 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
...
+        participants.push(participant)
+      }
+    }
 
-    if (result) {
-      const eventsMap: StringDictionary<ParticipantRelevantEvents> =
-        await ParticipantService.getRelevantEventsForParticipants(
+    if (total > 0) {
+      const participantData = []
+      const eventsMapPromises = participants.map(async participant => {
```
conundrumgirl, line 555: we already have functionality in getParticipants where we join participants with relevant events. Let's instead refactor this as a function that we can also use here

conundrumgirl (changes requested):
Looks good. A couple of refactoring suggestions to better comply with the DRY principle.

conundrumgirl (approved):

PR #251, opened 2021-07-09
Create download app landing page

leonli33:
Looks like:
Screen Shot 2021-07-09 at 10 10 44 AM


src/components/static/DownloadAppLandingPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <Box className={classes.downloadText}>
+        Download the <strong>Mobile Toolbox App</strong>
+      </Box>
+      <p className={classes.thankYouText}>
+        Thank you for participating.
+        <br></br>
+        <br></br>
+        Please select the store that works best on your smartphone:
+      </p>
+      <Box mt={3}>
```
conundrumgirl, line 56: I think it's more appropriate to you a link <a> rather than the button:
http://web-accessibility.carnegiemuseums.org/content/buttons/#:~:text=Clickable%20actions%3A%20Use%20buttons%20to,a%20link%20like%20a%20button.

conundrumgirl (changes requested):
One small change for button vs link otherwise great!

conundrumgirl (approved):

PR #257, opened 2021-07-13
Edit Customize App Page

leonli33:
1. Allow for partial editing of the fields
2. Display errors only when coming from the launch page.

### DIFF
--- a/src/components/studies/StudyBuilder.tsx
+++ b/src/components/studies/StudyBuilder.tsx
@@ -36,6 +36,8 @@ import Scheduler from './scheduler/Scheduler'
 import {StudySection} from './sections'
 import SessionCreator from './session-creator/SessionCreator'
 import StudyLeftNav from './StudyLeftNav'
+import constants from '../../types/constants'
+
 const subtitles: StringDictionary<string> = {
   description: 'Description',
   'team-settings': 'Team Settings',
@@ -119,11 +121,10 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
   ...otherProps
 }) => {
   const classes = useStyles()
-  let {id, section: _section} =
-    useParams<{
-      id: string
-      section: StudySection
-    }>()
+  let {id, section: _section} = useParams<{
+    id: string
+    section: StudySection
+  }>()
   const [section, setSection] = React.useState(_section)
   const [error, setError] = React.useState<string[]>([])
   const handleError = useErrorHandler()
@@ -208,7 +209,7 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
       if (e.statusCode === 401) {
         handleError(e)
       }
-      setError(e.message)
+      setError([e.message])
       window.scrollTo({
         top: 0,
         behavior: 'smooth',
@@ -520,15 +521,23 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
                           }}
                           onUpdate={(updatedStudy: Study) => {
                             setHasObjectChanged(true)
+                            if (updatedStudy.contacts) {
+                              for (const contact of updatedStudy.contacts) {
+                                if (contact.name === '') {
+                                  contact.name =
+                                    constants.constants.CONTACT_NAME_DEFAULT
+                                }
+                              }
+                            }
                             setData({
                               ...builderInfo,
                               study: updatedStudy,
                             })
-                            builderInfo.study = updatedStudy
                           }}
                           onError={(error: string) =>
                             setError(prev => [...prev, error])
-                          }>
+                          }
+                          errorMessage={error[0] || ''}>
                           {navButtons}
                         </AppDesign>
                       )}
--- a/src/components/studies/StudyBuilder.tsx
+++ b/src/components/studies/StudyBuilder.tsx
@@ -521,14 +521,6 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
                           }}
                           onUpdate={(updatedStudy: Study) => {
                             setHasObjectChanged(true)
-                            if (updatedStudy.contacts) {
-                              for (const contact of updatedStudy.contacts) {
-                                if (contact.name === '') {
-                                  contact.name =
-                                    constants.constants.CONTACT_NAME_DEFAULT
-                                }
-                              }
-                            }
                             setData({
                               ...builderInfo,
                               study: updatedStudy,
--- a/src/components/studies/StudyBuilder.tsx
+++ b/src/components/studies/StudyBuilder.tsx
@@ -536,8 +536,7 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
                           }}
                           onError={(error: string) =>
                             setError(prev => [...prev, error])
-                          }
-                          errorMessage={error[0] || ''}>
+                          }>
                           {navButtons}
                         </AppDesign>
                       )}
--- a/src/components/studies/app-design/AppDesign.tsx
+++ b/src/components/studies/app-design/AppDesign.tsx
@@ -365,7 +365,7 @@ const AppDesign: React.FunctionComponent<
     if (!contactLead?.position) {
       updatedErrorState.contactLeadPositonHasError = true
     }
-    if (irbRecord?.name === DEFAULT_CONTACT_NAME) {
+    if (!irbRecord || irbRecord?.name === DEFAULT_CONTACT_NAME) {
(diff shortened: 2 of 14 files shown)


src/components/studies/StudyBuilder.tsx
```diff
@@ -520,11 +521,18 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
                           }}
                           onUpdate={(updatedStudy: Study) => {
                             setHasObjectChanged(true)
+                            if (updatedStudy.contacts) {
```
conundrumgirl, line 524: I think this should happen in the AppDesign rather than in a parent component

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -279,11 +282,13 @@ const AppDesign: React.FunctionComponent<
   onUpdate,
   onSave,
   study,
-
   onError,
 }: AppDesignProps & StudyBuilderComponentProps) => {
   const handleError = useErrorHandler()
 
+  let query = useQuery()
+  const showError = query.get('from') === 'launch'
```
conundrumgirl, line 290: I think you don't need to check for a particular value e.g. launch. You know that 'launch' is the only place where it gets set. So you just want to check if it has some value. It also guards you agains misspelling e.g. 'lunch' value being passed

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -433,6 +509,11 @@ const AppDesign: React.FunctionComponent<
     }
   }, [])
 
+  const getContactName = (name: string | undefined) => {
+    if (name === DEFAULT_CONTACT_NAME) return ''
```
conundrumgirl, line 513: instead 
```
 if (name === DEFAULT_CONTACT_NAME) return ''
    return name ? name : ''

```
`return (name && name !== DEFAULT_CONTACT_NAME)? name : ''

src/components/studies/launch/LaunchAlerts.tsx
```diff
@@ -293,7 +304,14 @@ const StudyAlertComponent: React.FunctionComponent<
           </div>
           {error.isDismissable && (
             <Box className={classes.reviewIgnoreButtons}>
-              <Button href={SECTIONS[sectionIndex].path}>Review</Button>
+              <Button
+                href={
+                  SECTIONS[sectionIndex].path === 'customize'
```
conundrumgirl, line 309: I think it's better to handle it more generally. The other sections don't care about the 'from' parameter, but potentially they might so just doing     <Button href={`${SECTIONS[sectionIndex].path}?from=launch`} ... makes sense

src/components/studies/launch/LaunchAlerts.tsx
```diff
@@ -306,7 +324,11 @@ const StudyAlertComponent: React.FunctionComponent<
             <Button
               variant="contained"
               className={classes.mustReviewButton}
-              href={SECTIONS[sectionIndex].path}>
+              href={
+                SECTIONS[sectionIndex].path === 'customize'
```
conundrumgirl, line 328: same comment as above

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -316,6 +321,80 @@ const AppDesign: React.FunctionComponent<
     isIrbEmailValid: true,
   })
 
+  const [errorState, setErrorState] = React.useState({
```
conundrumgirl, line 324: you can have a type for errorState where all those props are optional. This way you can just initially set it to the empty object and don't need the lengthy set statements

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -316,6 +321,80 @@ const AppDesign: React.FunctionComponent<
...
+    contactLeadNameHasError: false,
+    contactLeadPositonHasError: false,
+    irbRecordNameHasError: false,
+    irbProtocolIdHasError: false,
+  })
+
+  useEffect(() => {
+    if (!showError) return
+    const updatedErrorState = {
+      studyTitleHasError: false,
```
conundrumgirl, line 338: same comment as above

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -640,6 +721,10 @@ const AppDesign: React.FunctionComponent<
                 }}
                 studyTitle={study.name || ''}
                 studySummaryBody={study.details || ''}
+                studyTitleHasError={showError && errorState.studyTitleHasError}
```
conundrumgirl, line 724: here and  below there is a bit of repetition.. What about making a function that takes the errorState property you are looking for? 
the fn would look something like this
```
function hasError(errorType: [your type can be a keyof of your error type]) {
return showError && errorState[errorType'
}
```
then you would just all studyTitleHasError={hasError('studyTitleHasError')}

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -316,6 +321,80 @@ const AppDesign: React.FunctionComponent<
...
+      studySummaryCopyHasError: false,
+      leadPINameHasError: false,
+      leadPIAffiliationHasError: false,
+      contactLeadNameHasError: false,
+      contactLeadPositonHasError: false,
+      irbRecordNameHasError: false,
+      irbProtocolIdHasError: false,
+    }
+    const contactLead = study.contacts?.find(el => el.role === 'study_support')
+    const principalInvestigator = study.contacts?.find(
```
conundrumgirl, line 348: is there a way to simplify/shorten this?  May be build an object that correlates the error with the property/value? 
It seems like you are applying similar logic to different properties There should be some way to take advantage of repeating pattern

conundrumgirl (changes requested):
I looks good, but I think there are opportunities for tightening up the code a bit.

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -316,6 +332,88 @@ const AppDesign: React.FunctionComponent<
...
+      !contactLead?.phone?.number || isInvalidPhone(generalContactPhoneNumber)
+    const irbRecordHasError =
+      !irbRecord?.phone?.number || isInvalidPhone(irbPhoneNumber)
+    setPhoneNumberErrorState({
+      isGeneralContactPhoneNumberValid: !generalContactPhoneError,
+      isIrbPhoneNumberValid: !irbRecordHasError,
+    })
+  }
+
+  const checkEmailError = (contactLead?: Contact, irbRecord?: Contact) => {
```
conundrumgirl, line 360: You can make this more generic as well  
Would something like this work? I think you can do this with the phone as well
```
const checkEmailError = (contact: Contact, error: keyof ErrorStateType) {
    const isError =v ! contact.email  || isInvalidEmail(contact.email)
     setEmailErrorState(...emailErrorState, [error]: isError)
}
```

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -316,6 +332,88 @@ const AppDesign: React.FunctionComponent<
...
+  }
+
+  useEffect(() => {
+    if (!showError) return
+    const updatedErrorState = {} as ErrorStateType
+    const contactLead = study.contacts?.find(el => el.role === 'study_support')
+    const principalInvestigator = study.contacts?.find(
+      c => c.role === 'principal_investigator'
+    )
+    const irbRecord = study.contacts?.find(el => el.role === 'irb')
```
conundrumgirl, line 387: Can 'getContact' function be used here?

src/components/studies/app-design/AppDesign.tsx
```diff
@@ -316,6 +332,88 @@ const AppDesign: React.FunctionComponent<
...
+    if (!isContactValid(principalInvestigator, 'name')) {
```
conundrumgirl, line 388: here and below how about 
` updatedErrorState.leadPINameHasError = !isContactValid(principalInvestigator, 'name') ` etc the you don't need the conditions

conundrumgirl (changes requested):

src/components/studies/app-design/AppDesign.tsx
leonli33: Hey, so in order for this to work, I wold have to call the checkEmailError function twice, one for the contactLead and another for the irbRecord. I was doing this before and was getting weird behavior because we are setting the same state twice in a very close time frame to each other, so some things were getting overwritten. A option would just be to return isError from the method and update it together back in the useEffect.

conundrumgirl (approved):

Issue #256, opened 2021-07-12
Search state when changing the tabs


conundrumgirl:
When you search for a participant and change the tab you end up in the state where you do have something in the searchbox but the tab results don't reflect it.
We need to do one of two things:
Either when we change the tab we should clear out the search OR when we change the tab we should perform the search with the value entered

PR #258, opened 2021-07-14
Clear search when tab is changed.

leonli33:
In the participant manager, when the tab is changed the search participant component is cleared.

Resolves: #256

### DIFF
--- a/src/components/studies/participants/ParticipantManager.tsx
+++ b/src/components/studies/participants/ParticipantManager.tsx
@@ -419,10 +419,7 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
     })
   const [isAllSelected, setIsAllSelected] = React.useState(false)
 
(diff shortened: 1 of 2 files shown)


src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -419,7 +419,10 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
     })
   const [isAllSelected, setIsAllSelected] = React.useState(false)
 
+  const [forceSearchReset, setForceSearchReset] = React.useState(false)
```
conundrumgirl, line 422: Do you need an additional state variable? If you need to clear when the tab is changed can you just pass the current tab as a property to the search control and have your reset logic based on that?

conundrumgirl (changes requested):

conundrumgirl (approved):

PR #263, opened 2021-07-16
Use alert for email and phone format error

leonli33:
Looks like:
Screen Shot 2021-07-16 at 2 42 27 PM


src/components/studies/app-design/IrbBoardContactSection.tsx
```diff
@@ -222,11 +231,17 @@ const IrbBoardContactSection: React.FunctionComponent<IrbBoardContactSectionProp
             titleText="Email*"
           />
           {!emailErrorState.isIrbEmailValid && (
-            <FormHelperText
-              id="ethics-bad-email-text"
-              className={classes.errorText}>
-              email should be in a valid format such as: example@placeholder.com
-            </FormHelperText>
+            <AlertWithText
```
conundrumgirl, line 234: can we have the icon be a part of AlertWithText?
If we have several possibilities we can have a variant that would add a class/image combo.
If it is challenging with StyledComponent (you might be able to use 'content' and :before in css to get it working) you can replace it w/  a regular component that uses styled component

conundrumgirl (changes requested):

src/components/studies/app-design/GeneralContactAndSupportSection.tsx
```diff
@@ -1,21 +1,27 @@
...
 import {ContactType} from './AppDesign'
+import AlertWithTextWrapper from '../../widgets/AlertWithTextWrapper'
 
 const useStyles = makeStyles(theme => ({
   firstFormElement: {
     marginTop: theme.spacing(2.5),
   },
   errorText: {
-    marginTop: theme.spacing(-0.5),
+    marginTop: theme.spacing(-1.5),
```
conundrumgirl, line 19: You are duplicating the class in another component. Can this class live in the AlertWithTextWrapper?

conundrumgirl (changes requested):

conundrumgirl (approved):

PR #265, opened 2021-07-19
Added placeholder participant manager screen

leonli33:
Looks Like:
Screen Shot 2021-07-19 at 11 12 20 AM


src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -618,6 +626,8 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
     setLoadingIndicators({isDownloading: false})
   }
 
+  const displayPlaceholderScreen =
```
leonli33, line 629: This was the only change I made to this file, besides line 647 - 654.

src/components/studies/participants/ParticipantManagerPlaceholder.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+import {makeStyles, Box} from '@material-ui/core'
+import RocketIcon from '../../../assets/participant-manager/participant_manager_rocket_icon.svg'
```
conundrumgirl, line 3: we already have /participants directory in the assets. let's keep all the images there. Also since you're doing that work can you please move  the images our /ParticipantManagerPageSelector into the subdirectory of /participants named /paging (needs to be created) and update the references

conundrumgirl (changes requested):
some changes regarding where to store images. Otherwise good!

conundrumgirl (approved):

PR #266, opened 2021-07-19
Remove failing tests

leonli33:


src/__tests__/components/studies/app-design/AppDesignTest.test.tsx
```diff
@@ -1,367 +1,369 @@
...
-  expect(
-    getById(appDesign as HTMLElement, 'general-contact-bad-phone-text')
-  ).not.toBeNull()
-  userEvent.clear(contactLeadPhoneNumber!)
-  userEvent.click(container)
-  expect(
-    getById(appDesign as HTMLElement, 'general-contact-bad-phone-text')
-  ).toBeNull()
-})
+// import {BrowserRouter} from 'react-router-dom'
```
conundrumgirl, line 10: Why `//` rather than `/*`?

conundrumgirl (changes requested):

src/__tests__/components/studies/app-design/AppDesignTest.test.tsx
leonli33: I highlighted everything and hit command + / , which uses `//`. Should i change it?

conundrumgirl: I suppose if there is an quick way to uncomment it - it's fine. In general // is for single line comments and /*. */ for blocks

PR #272, opened 2021-07-20
Create arc landing page

leonli33:
Looks Like:
Screen Shot 2021-07-20 at 3 26 53 PM


src/UnauthenticatedApp.tsx
```diff
@@ -7,17 +7,49 @@ import TopNav from './components/widgets/AppTopNav'
 import {setBodyClass} from './helpers/utility'
 import PublicRoutes from './routes_public'
 import constants from './types/constants'
+import ArcLogo from './assets/arc/arc_main_logo.svg'
```
conundrumgirl, line 10: do we need special directory for arc? I think it's just a logo. Let's keep it in the root of /assets

conundrumgirl (changes requested):
small tweak where to store the logo -- otherwise good!

conundrumgirl (approved):

PR #274, opened 2021-07-21
Fixed Access Settings Bug

### LINKED ISSUE #270: Bug adding user in access settings
when adding a new account roles do not get assigned correctly. This bug does not exist when updating the roles for the existing user
For testing purposes you can delete and add agendel_test@synapse.org
roles to access mapping is:
study_designer: study builder (developer who can configure)
study_coordinator: participant manager, adherence
org_admin: edit/add users

co-admin does not correspond to the specific role. Rather it has all of the 3 above roles

leonli33:
Fixed access settings bug and also made minor style changes.

Resolves #270


conundrumgirl (approved):
Great!

PR #277, opened 2021-07-23
Banner shown after study is published (read only pages.)

leonli33:
Looks Like: 
Screen Shot 2021-07-23 at 2 51 25 PM

Screen Shot 2021-07-23 at 2 50 58 PM

Screen Shot 2021-07-23 at 2 51 48 PM

Also, wrap session legend if there are too many for one line (requested by Lynn):
Screen Shot 2021-07-23 at 3 53 11 PM


src/components/studies/StudyBuilder.tsx
```diff
@@ -138,6 +141,58 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
...
+        bgColor: string
+        displayText: string[]
+        icon: string[]
+        textColor: string
+      }
+    >()
+  )
+
+  React.useEffect(() => {
+    const bannerMap = new Map()
```
conundrumgirl, line 157: Banner map is a constant so it would be better defined outside the components. May be even put it into a separate .ts file and import it?

src/components/studies/StudyBuilder.tsx
```diff
@@ -396,17 +451,33 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
   }
 
   const getBannerType = () => {
+    if (builderInfo?.study?.phase === 'in_flight') return 'live'
```
conundrumgirl, line 454: I think a switch statement might be appropriate here w/ 'completed' and 'analysis' be the default ?

src/components/studies/StudyBuilder.tsx
```diff
@@ -138,6 +141,58 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
...
+      icon: [NoEditIcon],
+      textColor: 'white',
+    })
+    bannerMap.set('withdrawn', {
+      bgColor: '#AA00FF',
+      displayText: ['The study was cancelled and cannot be edited.'],
+      icon: [NoEditIcon],
+      textColor: 'white',
+    })
+    setBannerTypes(bannerMap)
```
conundrumgirl, line 191: should the bannerTypes be in the state? are they dynamic? It would make sense to have bannerType in the state instead of /bannerTypes.get(bannerType)' function calls.

conundrumgirl (changes requested):
A couple of comments but looks good overall!

src/components/studies/StudyBuilder.tsx
```diff
@@ -138,6 +139,29 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
...
   const [open, setOpen] = React.useState(true)
   const [displayBanner, setDisplayBanner] = React.useState(false)
+  const [bannerType, setBannerType] = React.useState('')
+
+  React.useEffect(() => {
+    setBannerType(getBannerType())
+    if (getBannerType() !== 'success' && getBannerType() !== 'error') {
+      setDisplayBanner(true)
+    }
+  }, [])
```
conundrumgirl, line 149: I think it should depend on builderInfo?.study?.phase rather than []

src/components/studies/StudyBuilder.tsx
```diff
@@ -138,6 +139,29 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
...
+      case 'withdrawn':
+        return 'withdrawn'
+      case 'analysis':
+      case 'completed':
+        return 'completed'
+      default:
+        const errors = section === 'scheduler' ? schedulerErrors : error
+        return errors.length > 0 ? 'error' : 'success'
+    }
+  }
```
conundrumgirl, line 165: can we change it to take the phase (StudyPhase  type)  and section to make it more explicit rather than just using what's available in the component. It makes it more explicit what it depends on ./ We can also put it above the component

src/components/studies/StudyBuilder.tsx
```diff
@@ -395,18 +419,22 @@ const StudyBuilder: FunctionComponent<StudyBuilderProps> = ({
...
   return (
     <>
       <Box display="flex" bgcolor="#f7f7f7">
         <TopErrorBanner
+          backgroundColor={bannerInfo.bannerMap.get(bannerType)?.bgColor!}
+          textColor={bannerInfo.bannerMap.get(bannerType)?.textColor!}
           onClose={() => setDisplayBanner(false)}
           isVisible={displayBanner}
-          type={getBannerType()}></TopErrorBanner>
+          icon={bannerInfo.bannerMap.get(bannerType)?.icon[0]!}
```
conundrumgirl, line 431: can you use bannerType from the state?

conundrumgirl (approved):

PR #280, opened 2021-07-29
Cap session title at 18 characters

leonli33:
Screen Shot 2021-07-29 at 11 42 33 AM


conundrumgirl (changes requested):
To make it more general, let's replace shouldLimitCharacters prop w/ optional maxCharacters prop. If it's present -- limit to that number

conundrumgirl (approved):

PR #305, opened 2021-08-16
Added tooltip hover display to study types.

leonli33:
Looks like:
Screen Shot 2021-08-16 at 12 11 45 PM


src/components/studies/launch/AboutStudy.tsx
```diff
@@ -70,6 +72,44 @@ const suggestions = [
 ]
 
 const JOIN_TOKEN = '*'
+
+const StudyTypeLabel: React.FunctionComponent<{
+  type: 'observation' | 'validation' | 'intervention'
+  formLabelText: string
+}> = ({type, formLabelText}) => {
+  let label = (
```
conundrumgirl, line 80: I think 'switch statement is more appropriate here

src/components/widgets/InfoCircleWithToolTip.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
+import {Box, makeStyles, Tooltip} from '@material-ui/core'
+import ItalicI from '../../assets/italic_i_icon.svg'
+import {latoFont} from '../../style/theme'
+
+const useStyles = makeStyles(theme => ({
```
conundrumgirl, line 6: I think it makes sense here to take the color as a prop as opposed to hardcoding it to make it more reusable.

conundrumgirl (changes requested):
Good. A couple of comments. Please see the use of props w/ makeStyles

conundrumgirl (commented):
Nice!

PR #347, opened 2021-08-31
Cap scheduler duration at 104 weeks.

leonli33:


src/components/studies/scheduler/Duration.tsx
```diff
@@ -104,6 +106,15 @@ const Duration: React.FunctionComponent<
         {...props}
         id={numberLabel.replace(' ', '')}
         onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+          if (inputDurationCapInWeeks !== undefined) {
```
conundrumgirl, line 109: These checks should be a in a separate function which is called from the existing `changeValue` fn

conundrumgirl (changes requested):
Please change the limit to 5 years and provide the feedback for the user (red text?) if it's over that interval

conundrumgirl (approved):

PR #350, opened 2021-09-01
Fix participant issue in Participant Manager

leonli33:
Participants no longer move tabs when edited.

### DIFF
--- a/src/services/participants.service.ts
+++ b/src/services/participants.service.ts
@@ -591,16 +591,27 @@ async function updateParticipantNote(
   participantId: string,
   note?: string
 ) {
-  // update note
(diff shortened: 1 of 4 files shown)


src/services/participants.service.ts
```diff
@@ -585,22 +562,35 @@ async function addTestParticipant(
...
 }
 
-async function updateParticipantNote(
+async function updateParticipant(
   studyIdentifier: string,
   token: string,
   participantId: string,
-  note?: string
+  updatedFields: {
+    [Property in keyof ParticipantAccountSummary]?: ParticipantAccountSummary[Property]
```
conundrumgirl, line 570: Nice!

conundrumgirl (approved):

PR #351, opened 2021-09-01
Disable interval when occurrences is 1

leonli33:
Looks like:
Screen Shot 2021-09-01 at 3 23 26 PM

### DIFF
--- a/src/components/studies/scheduler/RepeatFrequency.tsx
+++ b/src/components/studies/scheduler/RepeatFrequency.tsx
@@ -22,10 +22,7 @@ const RepeatFrequency: React.FunctionComponent<RepeatFrequencyProps> = ({
   return (
     <SchedulingFormSection
       label={'Run this session every:'}
-      style={{
-        opacity: disabled ? 0.3 : 1,
-        pointerEvents: disabled ? 'none' : 'all',
-      }}>
+      disabled={disabled}>
       <FormControlLabel
         style={{
           marginLeft: '0',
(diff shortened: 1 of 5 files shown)


src/components/studies/scheduler/RepeatFrequency.tsx
```diff
@@ -18,14 +18,23 @@ const RepeatFrequency: React.FunctionComponent<RepeatFrequencyProps> = ({
   const label = occurrences
     ? ` for ${occurrences} times`
     : 'until the end of study'
-
+  const disabled = occurrences == 1
   return (
-    <SchedulingFormSection label={'Run this session every:'}>
+    <SchedulingFormSection
+      label={'Run this session every:'}
+      style={{
```
conundrumgirl, line 25: let's add a 'disabled' class here instead

src/components/studies/scheduler/SchedulableSingleSessionContainer.tsx
```diff
@@ -257,9 +257,16 @@ const SchedulableSingleSessionContainer: FunctionComponent<SchedulableSingleSess
           <Box className={classes.formSection}>
             <EndDate
               occurrences={schedulableSession.occurrences}
-              onChange={(occurrences: number | undefined) =>
-                updateSessionSchedule({...schedulableSession, occurrences})
-              }></EndDate>
+              onChange={(occurrences: number | undefined) => {
+                const newSessionSchedule = {
```
conundrumgirl, line 261: let's have a function instead of doing this inline

conundrumgirl (changes requested):

conundrumgirl (approved):

PR #353, opened 2021-09-02
Add timezone dropdown

leonli33:
Looks like:
Screen Shot 2021-09-02 at 4 00 08 PM
Screen Shot 2021-09-02 at 4 00 30 PM


src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -456,13 +456,22 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
   const updateParticipant = async (
     participantId: string,
     note: string,
-    customEvents: ParticipantEvent[]
+    customEvents: ParticipantEvent[],
```
conundrumgirl, line 459: can we use the same logic and just pass an object of options we want to update?

src/components/studies/participants/TimezoneDropdown.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const filtered = timezoneNames.filter(el => el.includes('America'))
+    const timezoneInfoArray: TimezoneInfoType[] = filtered.map(data => {
+      return {
+        label: data,
+        value: data,
+      }
+    })
+    return timezoneInfoArray
+  }
+  const timezones = React.useMemo(getAllTimezones, [])
```
conundrumgirl, line 25: good use of useMemo!

src/components/studies/participants/TimezoneDropdown.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <SaveBlackBorderDropdown
+      width="100%"
+      dropdown={timezones}
+      searchableOnChange={(event: string) => onValueChange(event)}
```
conundrumgirl, line 30: I think this can just be written as  `searchableOnChange={onValueChange}` please doblecheck. Also do we need to set the itemHeight?. Also curious what is the reason for empty function for `onChange`

src/components/studies/participants/grid/ParticipantTableGrid.tsx
```diff
@@ -266,6 +271,10 @@ function getDate(value: GridCellValue) {
   return value ? new Date(value as string).toLocaleDateString() : undefined
 }
 
+function getTimezone(value: GridCellValue) {
```
conundrumgirl, line 274: you don't need this function I think you can just do inline on ln380

src/components/studies/participants/grid/ParticipantTableGrid.tsx
```diff
@@ -141,11 +141,16 @@ const EditCell: FunctionComponent<{
         return result as ParticipantEvent[]
       }
 
+      const getClientTimeZone = (column: string) => {
```
conundrumgirl, line 144: I don't think you need a special function for getting clientTimeZone. You can modify getValString to take an optional default parameter

src/components/studies/participants/modify/EditParticipantForm.tsx
```diff
@@ -123,18 +127,25 @@ const EditParticipantForm: FunctionComponent<EditParticipantFormProps> = ({
...
-          <FormControl>
-            <SimpleTextLabel htmlFor="note">Notes</SimpleTextLabel>
-            <SimpleTextInput
-              value={note}
-              placeholder="comments"
-              onChange={e => setNotes(e.target.value)}
-              id="note"
-              multiline={true}
-              rows={5}
+          <Box width="375px" mb={3}>
```
conundrumgirl, line 130: Do we need to set the width?

src/components/widgets/BlackBorderDropdown.tsx
```diff
@@ -140,6 +154,37 @@ const SaveBlackBorderDropdown: React.FunctionComponent<
       ))}
     </Select>
   )
+
+  const searchableDropdown = (
+    <Box>
+      <Box style={{fontSize: '14px', fontFamily: poppinsFont}}>
+        Participant Time Zone
```
conundrumgirl, line 161: This is a generic control. It should not have things specific to a certain usecase

conundrumgirl (changes requested):

src/components/studies/participants/modify/EditParticipantForm.tsx
leonli33: Yea, so if there is no set width then the dropdown would span the entire length of the modal: 
Screen Shot 2021-09-03 at 2 55 20 PM

PR #356, opened 2021-09-08
Added no participants screen

leonli33:
Looks like:
Screen Shot 2021-09-08 at 11 15 42 AM
Screen Shot 2021-09-08 at 11 15 56 AM


src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -729,38 +782,40 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
...
-                          {!loadingIndicators.isDownloading ? (
-                            'StudyParticipants.csv'
-                          ) : (
-                            <CircularProgress size={24} />
-                          )}
-                        </>
-                      </ParticipantDownloadTrigger>
-
-                      {tab !== 'WITHDRAWN' && (
+                      {!displayNoParticipantsPage && (
```
conundrumgirl, line 785: I don't think we should have those repeated checks. We should replace the entire area with the 'no participants' screen. Also, let's have it in a separate file.

conundrumgirl (changes requested):
Ideally the changes to the actual ParticipantManager should be minimal

conundrumgirl (approved):

PR #487, opened 2022-01-27
style(scheduler): change edit icon color when hovering

yjyoo773:
Removed gray background when hovering over pencil icon and have it change colors to white.
https://sagebionetworks.jira.com/browse/MTB-403

### DIFF
--- a/src/components/studies/scheduler/SchedulableSingleSessionContainer.tsx
+++ b/src/components/studies/scheduler/SchedulableSingleSessionContainer.tsx
@@ -1,8 +1,7 @@
 import AlertIcon from '@assets/alert_icon.svg'
-import {ReactComponent as EditIcon} from '@assets/edit_pencil_red.svg'
 import InfoCircleWithToolTip from '@components/widgets/InfoCircleWithToolTip'
 import {AlertWithText, BlueButton} from '@components/widgets/StyledComponents'
-import {Box, IconButton, makeStyles, Switch, Tooltip} from '@material-ui/core'
+import {Box, makeStyles, Switch} from '@material-ui/core'
 import {DEFAULT_NOTIFICATION} from '@services/schedule.service'
 import {latoFont, ThemeType} from '@style/theme'
 import {
@@ -279,6 +278,7 @@ const SchedulableSingleSessionContainer: FunctionComponent<SchedulableSingleSess
         <form noValidate autoComplete="off">
           <Box className={classes.formSection}>
             <StartDate
+              onOpenEventsEditor={onOpenEventsEditor}
               startEventId={
                 _.isEmpty(studySession.studyBurstIds)
                   ? _.first(schedulableSession.startEventIds)!
@@ -300,12 +300,6 @@ const SchedulableSingleSessionContainer: FunctionComponent<SchedulableSingleSess
               onChangeDelay={(delay: string | undefined) => {
                 updateSessionSchedule({...schedulableSession, delay})
               }}>
-              <Tooltip title="Edit Custom Event">
-                <IconButton onClick={() => onOpenEventsEditor()}>
-                  {' '}
-                  <EditIcon />
-                </IconButton>
-              </Tooltip>
               <InfoCircleWithToolTip
                 tooltipDescription={
                   <span>
--- a/src/components/studies/scheduler/StartDate.tsx
+++ b/src/components/studies/scheduler/StartDate.tsx
@@ -1,18 +1,22 @@
 import {
   FormControl,
   FormGroup,
+  IconButton,
   makeStyles,
   MenuItem,
   Radio,
   RadioGroup,
   Select,
+  Tooltip,
 } from '@material-ui/core'
 import {JOINED_EVENT_ID} from '@services/event.service'
 import constants from '@typedefs/constants'
 import {HDWMEnum, SchedulingEvent} from '@typedefs/scheduling'
-import React from 'react'
+import React, {FunctionComponent} from 'react'
 import Duration from './Duration'
 import SchedulingFormSection from './SchedulingFormSection'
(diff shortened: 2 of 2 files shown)


src/components/studies/participantHooks.ts
```diff
@@ -0,0 +1,90 @@
...
+            refetchOnWindowFocus: false,
+        }
+    )
+}
+
+export const useUpdateParticipantInList = () => {
+    const {token} = useUserSessionDataState()
+    const queryClient = useQueryClient()
+
+    const update = async(props: {
```
conundrumgirl, line 41: We already have a hook to update the events so this hook should only be updating the fields on participant, Also I don't think we need the page (current/size/tab) information. I think we can invalidate all of the  PARTICIPANT_KEYS.list(props.studyId) not just for the particular page.

src/components/studies/participantHooks.ts
```diff
@@ -0,0 +1,90 @@
+import { useUserSessionDataState } from "@helpers/AuthContext"
+import ParticipantUtility, {ParticipantData} from "./participants/participantUtility"
+import { ExtendedError, ExtendedParticipantAccountSummary, ParticipantAccountSummary, ParticipantActivityType, ParticipantEvent } from "@typedefs/types"
+import { useMutation, useQuery, useQueryClient } from "react-query"
+import ParticipantService from "@services/participants.service"
+import EventService from "@services/event.service"
+
+export const PARTICIPANT_KEYS = {
+    all: ['participants'] as const,
+    list: (studyId: string | undefined, currentPage: number = 0, pageSize: number = 0, tab: ParticipantActivityType = 'ACTIVE', searchValue = undefined) =>
```
conundrumgirl, line 10: I don't think ParticipantActivityType should be optional. The reason is that when we invalidate this query we want to  make sure that the right one gets invalidated. e.g. if we withdraw someone we want to make sure that we make the 'WITHDRAWN' query stale

src/components/studies/participantHooks.ts
```diff
@@ -0,0 +1,90 @@
...
+    detail: (studyId: string, userId: string) =>
+      [...PARTICIPANT_KEYS.list(studyId), userId] as const,
+}
+
+export const useParticipants = (
+    studyId: string | undefined, 
+    currentPage: number, 
+    pageSize: number,
+    tab: ParticipantActivityType, 
+    searchValue?: string
```
conundrumgirl, line 22: what does searchValue do?

src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -206,34 +208,111 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
...
+  const {data: data, error : participantError, status: status} = useParticipants(study?.identifier,currentPage,pageSize,tab)
+  
+  const {    
+    mutate,
+    isSuccess,
+    isError,
+    isLoading: isParticipantUpdating,
+  } = useUpdateParticipantInList()
+  
+  const [curStatus, setCurStatus] = React.useState<RequestStatus>('PENDING')
```
conundrumgirl, line 232: I don't think we need  a state variable here. Can we just use the `status` from the `useParticipants`?

src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -206,34 +208,111 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
...
+  // } = useAsync<ParticipantData>({
+  //   status: 'PENDING',
+  //   data: null,
+  // })
+  // console.log("DATA",data?.items)
+
+  const {data: data, error : participantError, status: status} = useParticipants(study?.identifier,currentPage,pageSize,tab)
+  
+  const {    
+    mutate,
```
conundrumgirl, line 226: I don't think is an asynchronous call

src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -206,34 +208,111 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
...
+          userId, 
+          note,
           currentPage,
           pageSize,
-          tab,
-          token!
-        )
-      )
+          tab})
+        setRefreshParticipantsToggle(prev => !prev)
```
conundrumgirl, line 263: we should not need the refreshParticipantsToggle any more

src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -206,34 +208,111 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
...
+        return
+      case 'DELETE':
+        setLoadingIndicators(_ => ({isDeleting: true}))
+
+        for(let i = 0; i < selectedParticipantIds[tab!].length; i++){
+          await mutate({
```
conundrumgirl, line 269: can we change the mutation function to take the userId array  instead?

src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -206,34 +208,111 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
...
+            currentPage,
+            pageSize,
+            tab})
+        }
+        setLoadingIndicators(_ => ({isDeleting: false}))
+        setDialogState({dialogOpenRemove: false, dialogOpenSMS: false})
+        setRefreshParticipantsToggle(prev => !prev)
+        return
+      case 'UPDATE':
+        await mutate({
```
conundrumgirl, line 282: I don't think those awaits will work. You can console.log to confirm. You can specify success/error callbacks (preferable) or use mutateAsync (if that does not work)

conundrumgirl (changes requested):

conundrumgirl (approved):

PR #505, opened 2022-02-08
fix(participant): can add custom events to participants that were cre…

yjyoo773:
…ated without custom event

https://sagebionetworks.jira.com/browse/MTB-428


conundrumgirl:
no changes

PR #510, opened 2022-02-09
fix(pagination): change pagination index from 1 to 0

yjyoo773:
https://sagebionetworks.jira.com/browse/MTB-430

### DIFF
--- a/src/components/adherence/participants/AdherenceParticipants.tsx
+++ b/src/components/adherence/participants/AdherenceParticipants.tsx
@@ -65,7 +65,7 @@ const AdherenceParticipants: FunctionComponent<AdherenceParticipantsProps> =
 
     const {data: fullAdherenceWeeklyReport} = useAdherenceForWeek(
       studyId,
-      currentPage + 1,
+      currentPage,
       pageSize,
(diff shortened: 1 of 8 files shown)


src/components/adherence/participants/AdherenceParticipants.tsx
```diff
@@ -52,19 +52,20 @@ const AdherenceParticipants: FunctionComponent<AdherenceParticipantsProps> =
...
+
     const [pageSize, setPageSize] = React.useState(5)
     const [sessions, setSessions] = React.useState<SessionDisplayInfo[]>([])
 
     const {
       data: adherenceWeeklyReport,
       status: adhStatus,
       error: adhError,
-    } = useAdherenceForWeek(studyId, currentPage, pageSize, adherenceParams)
+    } = useAdherenceForWeek(studyId, currentPage + 1, pageSize, adherenceParams)
```
conundrumgirl, line 64: can we update adherence hook to also use 0-indexing? We will also need to update the service in that case

conundrumgirl (changes requested):

PR #518, opened 2022-02-14
feat(PM): create error handling alerts for features in PM

yjyoo773:
Create alert banner when error occurs while adding, withdrawing, editing, deleting, getting, or searching participants in participant manager.  

https://sagebionetworks.jira.com/browse/MTB-439

### DIFF
--- a/src/components/studies/participants/ParticipantDeleteModal.tsx
+++ b/src/components/studies/participants/ParticipantDeleteModal.tsx
@@ -1,9 +1,10 @@
-import { FunctionComponent } from "react";
+import React, { FunctionComponent } from "react";
 import {
     Dialog,
     DialogActions,
     DialogContent,
   } from '@material-ui/core'
+import Alert from '@material-ui/lab/Alert'
 import {ReactComponent as DeleteIcon} from '@assets/trash.svg'
 import DialogTitleWithClose from '@components/widgets/DialogTitleWithClose'
 import {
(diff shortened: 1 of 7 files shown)


src/components/studies/participants/ParticipantDeleteModal.tsx
```diff
@@ -73,33 +74,37 @@ const ParticipantDeleteModal: FunctionComponent<ParticipantDeleteModalProps> = (
     } = useStudy(studyId)
 
     const {data} = useParticipants(study?.identifier, currentPage, pageSize, tab,)
-    const {mutateAsync} = useUpdateParticipantInList()
+    const {mutateAsync, error: deleteParticipantError, isSuccess} = useUpdateParticipantInList()
     const invalidateParticipants = useInvalidateParticipants()
-      
+    const [error, setError] = React.useState<Error>()
+    React.useEffect(()=>{
```
conundrumgirl, line 80: do we need `error` state variable? Can we just use deleteParticipantError?

src/components/studies/participants/ParticipantManager.tsx
```diff
@@ -268,6 +266,10 @@ const ParticipantManager: FunctionComponent<ParticipantManagerProps> = () => {
     setIsAllSelected(false)
   }
 
+  React.useEffect(()=>{
+    if(participantError) setError(participantError)
```
conundrumgirl, line 270: same question about using participantError directly

src/components/studies/participants/add/AddSingleParticipant.tsx
```diff
@@ -58,9 +58,13 @@ const AddSingleParticipant: FunctionComponent<AddSingleParticipantProps> = ({
       externalId: '',
     }
   )
-  const [error, setError] = React.useState('')
+  const [error, setError] = React.useState<Error>()
+  const {isLoading, error: participantAddError, mutateAsync} = useAddParticipant()
+
+  React.useEffect(()=>{
```
conundrumgirl, line 64: same q. as above

conundrumgirl (commented):
a couple of questions about having a separate error state var. rather than using the error returned from the hook

src/components/studies/participants/ParticipantDeleteModal.tsx
yjyoo773: I had it set up this way because when trying to get the message in JSX, deleteParticipantError was acknowledged as an unknown object and to remove the banner when the modal was closed. Otherwise the alert would still be there when opening the modal.

conundrumgirl: got it

conundrumgirl (approved):

PR #559, opened 2022-04-06
fix(scheduler): disable update burst button when not using bursts

yjyoo773:


src/components/studies/scheduler/Scheduler.tsx
```diff
@@ -809,6 +813,8 @@ const Scheduler: React.FunctionComponent<SchedulerProps> = ({
         </DialogTitle>
         <DialogContent style={{padding: 0}}>
           <ConfigureBurstTab
+            hasBursts={hasBursts}
+            setHasBursts={setHasBursts}
```
conundrumgirl, line 817: I would rename setHasBursts to onSetHasBursts to make clear that it's a callback Fn. Otherwise -- looks good

conundrumgirl (changes requested):

PR #565, opened 2022-04-11
fix(scheduler): use labels as placeholders and hide label when value …

yjyoo773:
…is undefined

### DIFF
--- a/src/components/studies/scheduler/Duration.tsx
+++ b/src/components/studies/scheduler/Duration.tsx
@@ -33,6 +33,7 @@ export interface DurationProps {
   disabled?: boolean
   isShowClear?: boolean
   placeHolder? :string
+  selectWidth? : number
 }
 
 const Duration: React.FunctionComponent<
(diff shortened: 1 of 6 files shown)


src/components/studies/scheduler/Duration.tsx
```diff
@@ -127,12 +129,15 @@ const Duration: React.FunctionComponent<
...
         {...props}
         value={unt}
+        label={placeHolder}
         sourceData={unitData}
         id={unitLabel.replace(' ', '')}
         onChange={e =>
           changeValue(num, e.target.value as moment.unitOfTime.Base)
         }
-        style={isIntro ? {width: '100px'} : undefined}></SelectWithEnum>
+        style={{width: '100px'}}>
```
conundrumgirl, line 138: Instead of hard setting it, let's have it as an optional prop w/ a default value of 100px. That way we can reuse it and give it different width if we need to

conundrumgirl (approved):

PR #571, opened 2022-04-21
fix(scheduler): style burst button when inactive and disable only whe…

yjyoo773:
…n burst design off and nothing to save

### DIFF
--- a/src/components/studies/scheduler/Scheduler.tsx
+++ b/src/components/studies/scheduler/Scheduler.tsx
@@ -605,7 +605,7 @@ const Scheduler: React.FunctionComponent<SchedulerProps> = ({
                 disabled={isScheduleDefault(schedule) && !hasBeenSaved}
                 className={classes.burstButton}
                 onClick={() => setOpenModal('BURSTS')}>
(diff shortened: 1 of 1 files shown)


src/components/studies/scheduler/Scheduler.tsx
```diff
@@ -827,7 +830,8 @@ const Scheduler: React.FunctionComponent<SchedulerProps> = ({
           </DialogButtonSecondary>
 
           <DialogButtonPrimary
-            disabled={!!!hasBursts}
+            disabled={!!!hasBursts && schedule.studyBursts?.length===0}
```
conundrumgirl, line 833: why do we need this second check? can it be !!!hasBursts && schedule.studyBursts?.length>0? If so, there is probably a bug

yjyoo773: We can set schedule.studyBursts?.length < 1 instead of schedule.studyBursts?.length===0. As for the double check, the hasBursts is a boolean which is triggered by the toggle. Without the second check the user would not be able to save when they no longer want to do a burst.

PR #692, opened 2023-01-30
  add tracking to mtb studies

conundrumgirl:
https://sagebionetworks.jira.com/browse/MTB-647


src/App.tsx
```diff
@@ -45,38 +47,40 @@ function App() {
   }, [])
 
   return (
-    <StyledEngineProvider injectFirst>
```
conundrumgirl, line 48: just a refactor so that app is wrapped in Router

PR #703, opened 2023-02-15
Researcher UI for Alerts

hallieswan:
Work in progress for https://sagebionetworks.jira.com/browse/MTB-652

### DIFF
--- a/src/components/adherence/summary/AdherenceAlerts.tsx
+++ b/src/components/adherence/summary/AdherenceAlerts.tsx
@@ -191,9 +191,6 @@ const AlertMenu: FunctionComponent<{
       studyId: studyId,
       alertIds: [alert.id],
       action: action,
-      categories: checked.join(),
-      currentPage: currentPage,
-      pageSize: pageSize,
     })
   }
 
@@ -242,8 +239,7 @@ const AlertMenu: FunctionComponent<{
         onCancel={closeConfirmationDialog}
         onConfirm={() => {
           closeConfirmationDialog()
-          console.log('DELETE THE ALERT')
-          // onUpdateAdherenceAlert('DELETE')
+          onUpdateAdherenceAlert('DELETE')
         }}>
         <div>
           <p>
--- a/src/components/adherence/summary/AdherenceAlerts.tsx
+++ b/src/components/adherence/summary/AdherenceAlerts.tsx
@@ -1,143 +1,153 @@
-import {ReactComponent as AdherenceRed} from '@assets/adherence/adherence_icon_red.svg'
-import {ReactComponent as CalendarBlue} from '@assets/adherence/calendar_blue.svg'
-import {ReactComponent as PersonIcon} from '@assets/adherence/person_icon.svg'
-import {ReactComponent as CheckEmpty} from '@assets/adherence/round_check_empty.svg'
-import {ReactComponent as CheckGreen} from '@assets/adherence/round_check_green.svg'
 import {useAdherenceAlerts} from '@components/studies/adherenceHooks'
 import Loader from '@components/widgets/Loader'
-import CommentIcon from '@mui/icons-material/Comment'
-import {Box, Checkbox, styled} from '@mui/material'
-import IconButton from '@mui/material/IconButton'
-import List from '@mui/material/List'
-import ListItem from '@mui/material/ListItem'
-import ListItemButton from '@mui/material/ListItemButton'
-import ListItemIcon from '@mui/material/ListItemIcon'
-import ListItemText from '@mui/material/ListItemText'
-import {latoFont} from '@style/theme'
+import TablePagination from '@components/widgets/pagination/TablePagination'
+import {
+  Box,
+  Checkbox,
+  FormControlLabel,
+  FormGroup,
+  Table,
+  TableRow,
+  TableBody,
+  IconButton,
+  Typography,
+  Avatar,
+  styled,
+} from '@mui/material'
+import React, {FunctionComponent, ReactElement} from 'react'
+import {AdherenceAlert, AdherenceAlertCategory} from '@typedefs/types'
+import {theme} from '@style/theme'
 import dayjs from 'dayjs'
-import React, {FunctionComponent} from 'react'
+import {BorderedTableCell, StyledLink} from '../../widgets/StyledComponents'
+import {DateRange, PersonAddAlt, QuestionMark, TrendingDown} from '@mui/icons-material'
+import MoreVertIcon from '@mui/icons-material/MoreVert'
+import Menu from '@mui/material/Menu'
+import MenuItem from '@mui/material/MenuItem'
+import NotificationsNoneTwoToneIcon from '@mui/icons-material/NotificationsNoneTwoTone'
+import ParticipantService from '@services/participants.service'
+import PublishedWithChangesRoundedIcon from '@mui/icons-material/PublishedWithChangesRounded'
 
 type AdherenceAlertsProps = {
   studyId: string
 }
 
-type AlertType = 'enrollment' | 'adherence' | 'upcoming'
+const CheckboxBox = styled(Box)(({theme}) => ({
+  padding: theme.spacing(1, 2),
+  display: 'flex',
+  mt: 0,
+  mb: 1,
+  background: '#F1F3F5',
+}))
 
-type AdhAlert = {
-  type: AlertType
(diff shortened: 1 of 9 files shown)


src/types/constants.ts
```diff
@@ -116,6 +116,9 @@ const constants = {
 
   endpoints: {
     adherenceAlerts: '/v5/studies/:studyId/alerts',
+    adherenceAlertsMarkRead: '/v5/studies/:studyId/alerts/read',
```
conundrumgirl, line 119: I don't think we need these 2 endpoints

src/services/adherence.service.ts
```diff
@@ -142,8 +142,39 @@ async function getAdherenceAlerts(
   return result.data
 }
 
+async function markAdherenceAlertsAsRead(
```
conundrumgirl, line 145: I would have a single function markAdherenceAlerts that takes read/unread as a parameter

src/services/adherence.service.ts
```diff
@@ -142,8 +142,39 @@ async function getAdherenceAlerts(
...
+  studyId: string,
+  alertIds: string[],
+  token: string
+): Promise<any> {
+  const endpoint = constants.endpoints.adherenceAlertsMarkRead.replace(':studyId', studyId)
```
conundrumgirl, line 150: here I would use adherenceAlerts endpoint constant replace studyId as you are doing and then concatenate '/unread' or '/read' depending on the param

src/services/adherence.service.ts
```diff
@@ -142,8 +142,39 @@ async function getAdherenceAlerts(
...
+async function markAdherenceAlertsAsUnread(
+  studyId: string,
+  alertIds: string[],
+  token: string
+): Promise<any> {
+  const endpoint = constants.endpoints.adherenceAlertsMarkUnread.replace(':studyId', studyId)
+  const result = await Utility.callEndpoint(
+    endpoint,
+    'POST',
+    {studyId: studyId, alertIds: alertIds},
```
conundrumgirl, line 169: it's confusing in the documentation but I think you only need alertIds since you are passing a studyId in the url

src/components/studies/adherenceHooks.ts
```diff
@@ -88,6 +88,43 @@ export const useAdherenceAlerts = (
...
+  const queryClient = useQueryClient()
+
+  // update alerts status on the server
+  const update = async (props: {
+    studyId: string, 
+    alertIds: string[], 
+    action: 'READ' | 'UNREAD' | 'DELETE'
+  }): Promise<any> => {
+    const {studyId, alertIds} = props
+    switch(props.action) {
```
conundrumgirl, line 102: since we are going to have a single function in the service -- you don't need this switch - just pass the action param to the service function

src/components/adherence/summary/AdherenceAlerts.tsx
```diff
@@ -1,143 +1,212 @@
...
+import ParticipantService from '@services/participants.service'
+import PublishedWithChangesRoundedIcon from '@mui/icons-material/PublishedWithChangesRounded'
+import PersonPinCircleOutlinedIcon from '@mui/icons-material/PersonPinCircleOutlined'
 
 type AdherenceAlertsProps = {
   studyId: string
 }
 
-type AlertType = 'enrollment' | 'adherence' | 'upcoming'
+const CheckboxBox = styled(Box)(({theme}) => ({
```
conundrumgirl, line 35: We we need to style this checkbox? It's not a standard one we use?

src/components/adherence/summary/AdherenceAlerts.tsx
```diff
@@ -1,143 +1,212 @@
...
-    timestamp: new Date(),
-    text: 'has and upcoming Burst in two weeks',
-  },
-  {
-    type: 'enrollment',
-    participantId: '32311432',
-    timestamp: new Date(),
-    text: 'has enrolled',
-  },
+const ALERT_CATEGORY: {label: string; value: AdherenceAlertCategory}[] = [
```
conundrumgirl, line 43: I would have a single object that would be something like 
```
const ALERTS = {
new_enrollment: { label: 'New Enrollment', icon:  <PersonAddAlt />, text: 'has newly enrolled'},
low_adherence: { label: 'Adherence', icon:   <TrendingDown />, text: 'has adherence below [threshold]}`,
....
}
```

then you'd just do alertCategoryToLabel = ALERTS[whateveryourtypeis]. Same with images. for text you can do the same and then you can do `if(whateveryourtypeis ===  'low_adherence') {text = text.replace([threshold], JSON.parse(data).adherenceThreshold)}

src/components/studies/adherenceHooks.ts
```diff
@@ -88,6 +88,43 @@ export const useAdherenceAlerts = (
...
+
+  // invalidate alerts data
+  const mutation = useMutation(update, {
+    onMutate: async props => {
+      queryClient.cancelQueries(ADHERENCE_KEYS.alerts(props.studyId))
+    },
+    onError: (err, variables, context) => {
+      console.log(err, variables, context)
+    }, 
+    onSettled: async (data, error, props) => {
```
conundrumgirl, line 121: Nice!

conundrumgirl (approved):

src/components/studies/adherenceHooks.ts
```diff
@@ -73,13 +76,13 @@ export const useAdherenceAlerts = (
...
+  currentPage = 0
 ) => {
-  const cat = categories.join()
+  const cat = categories.sort().join()
   const {token} = useUserSessionDataState()
   return useQuery<{items: AdherenceAlert[]; total: number}, ExtendedError>(
-    ADHERENCE_KEYS.alerts(studyId, cat, offsetBy, pageSize),
-    () => AdherenceService.getAdherenceAlerts(studyId, categories, pageSize, offsetBy, token!),
+    ADHERENCE_KEYS.alerts(studyId, cat, currentPage, pageSize),
+    () => AdherenceService.getAdherenceAlerts(studyId, categories, pageSize, currentPage, token!),
```
nickgros, line 85: FYI, I see the same bug that we just had to fix in Synapse. If the `token` changes, neither `invalidateQueries` nor `resetQueries` would result in the query function using the new token. 

We added the token to the query key, but a better solution might be to have the AdherenceService pull the token from a stateful store (which is what we will probably end up switching to in the future).

src/components/studies/adherenceHooks.ts
```diff
@@ -88,6 +91,34 @@ export const useAdherenceAlerts = (
...
+  // invalidate alerts data
+  const mutation = useMutation(update, {
+    onMutate: async props => {
+      queryClient.cancelQueries(ADHERENCE_KEYS.alertsAll(props.studyId))
+    },
+    onError: (err, variables, context) => {
+      console.log(err, variables, context)
+    }, 
+    onSettled: async (data, error, props) => {
+      queryClient.invalidateQueries({queryKey: ADHERENCE_KEYS.alertsAll(props.studyId), refetchActive: true, refetchInactive: true})
```
nickgros, line 116: I would think TypeScript would catch this if it was a problem, but this looks like the react-query v4 signature when you're using react-query v3.

See [QueryClient.invalidateQueries ](https://react-query-v3.tanstack.com/reference/QueryClient#queryclientinvalidatequeries)

Maybe after this change you could remove these extra parameters?

Also, if you want the cache data to be dropped (so `useAdherenceAlerts` would immediately enter the `loading` state, assuming there isn't a default initial data value), you'd want to use [resetQueries](https://react-query-v3.tanstack.com/reference/QueryClient#queryclientresetqueries)

```suggestion
      queryClient.invalidateQueries(ADHERENCE_KEYS.alertsAll(props.studyId), { refetchActive: true, refetchInactive: true})
```

src/components/studies/adherenceHooks.ts
conundrumgirl: @nickgros, this makes sense, but I don't think it would be an issue in MTB since I can't think of the scenario where the token would change without the app reinitializing, since login involves redirection away from the app itself

src/components/studies/adherenceHooks.ts
conundrumgirl: @nickgros - you are talking about `{queryKey:..` YES! I didn't catch that and strangely neither did TS. Good call. It is strange that it still seemed to work. I will try the 'reset'

PR #706, opened 2023-02-23
Add styled verify now banner

hallieswan:
https://sagebionetworks.jira.com/browse/DHP-234

### DIFF
--- a/src/components/studies/StudyList.tsx
+++ b/src/components/studies/StudyList.tsx
@@ -4,6 +4,7 @@ import ConfirmationDialog, {ConfirmationDialogType} from '@components/widgets/Co
 import Loader from '@components/widgets/Loader'
 import {useUserSessionDataState} from '@helpers/AuthContext'
 import Utility from '@helpers/utility'
+import InfoTwoToneIcon from '@mui/icons-material/InfoTwoTone'
 import {Box, Button, Container, Menu, MenuItem, styled, Typography} from '@mui/material'
 import Link from '@mui/material/Link'
 import StudyService from '@services/study.service'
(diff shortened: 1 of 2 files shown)


src/components/studies/StudyList.tsx
```diff
@@ -176,6 +178,59 @@ const StudySublist: FunctionComponent<StudySublistProps> = ({
   )
 }
 
+const VerifyBanner: FunctionComponent<{isVerified: boolean | undefined}> = ({isVerified}) => {
+  const [displayVerifyBanner, setDisplayVerifyBanner] = React.useState<boolean>(!isVerified)
+
+  if (typeof isVerified === undefined) {
```
conundrumgirl, line 184: You can display this conditionally in a parent (below) only if verified !== true. And then you don't need 'isVerified' param

src/components/studies/StudyList.tsx
```diff
@@ -482,6 +533,7 @@ const StudyList: FunctionComponent<StudyListProps> = () => {
           </Container>
         </Box>
       </Box>
+      <VerifyBanner isVerified={isVerified} />
```
conundrumgirl, line 536: `{isVerified !== true && <VerifyBanner /> }`

src/components/widgets/AlertBanner.tsx
```diff
@@ -83,14 +87,21 @@ const AlertBanner: React.FunctionComponent<AlertBannerProps> = ({
       onClose={() => onClose()}
       severity="error"
       style={{color: textColor, backgroundColor: backgroundColor}}
+      sx={{
```
conundrumgirl, line 90: nice!

conundrumgirl (changes requested):
a couple of small things

src/components/studies/StudyList.tsx
hallieswan: Oh got it, makes sense! Done in next commit

conundrumgirl (approved):

PR #710, opened 2023-03-03
Bug fix for reminder notification and intro info 

conundrumgirl:


src/components/studies/scheduler/IntroInfo.tsx
```diff
@@ -147,21 +147,17 @@ const IntroInfo: React.FunctionComponent<IntroInfoProps> = ({id, onShowFeedback,
           duration,
         }
       }
-
-      mutateSchedule(
-        {
+      try {
```
conundrumgirl, line 150: changing mutate to async with await and try/catch blocks. The redirect only happens when everything is compelted

hallieswan (approved):

PR #720, opened 2023-03-15
Add session scheduler tests - notifications

hallieswan:
- don't allow initial notification to be deleted if a follow-up notification exists
- re-calculate multiday reminder offset when windowStartTime is changed
- add session scheduler tests - notifications

### DIFF
--- a/src/components/studies/scheduler/NotificationTime.tsx
+++ b/src/components/studies/scheduler/NotificationTime.tsx
@@ -114,6 +114,32 @@ const NotificationTime: React.FunctionComponent<NotificationTimeProps> = ({
     }
   }, [offset, isMultiday])
 
+  // Hook to track previous value of a prop, from: https://stackoverflow.com/a/57706747
+  const usePrevious = <T extends unknown>(value: T): T | undefined => {
+    const ref = React.useRef<T>()
+    React.useEffect(() => {
+      ref.current = value
+    })
+    return ref.current
+  }
+
+  /* Re-calculate offset when windowStartTime changes. 
+  Track prevWindowStartTime so can tell when windowStartTime has actually changed, as opposed to 
+  just when the offset is updated, which updates the NotificationWindow key and treats windowStarTime as changed. 
+  See: https://stackoverflow.com/a/60979832 */
+  const prevWindowStartTime = usePrevious(windowStartTime)
+  React.useEffect(() => {
+    if (
+      isFollowUp &&
+      isMultiday &&
+      prevWindowStartTime !== undefined &&
+      windowStartTime !== undefined &&
+      prevWindowStartTime !== windowStartTime
+    ) {
+      changeMultidayOffset({days: daysForMultidayOffset, time: timeForMultidayOffset})
+    }
+  }, [windowStartTime])
+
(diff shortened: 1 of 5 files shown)


src/components/studies/scheduler/SchedulableSingleSessionContainer.test.tsx
```diff
@@ -555,80 +565,759 @@ describe('SchedulableSingleSessionContainer', () => {
       expect(addButton).toBeDisabled()
     })
   })
-})
 
-test.skip('renders and changes reminder notification type', async () => {
```
hallieswan, line 560: The diff is a bit misleading here -- the previous notification tests were moved into the session notification subsection and are still at the bottom of this set of tests

src/components/studies/scheduler/SchedulableSingleSessionContainer.tsx
```diff
@@ -216,6 +216,14 @@ const SchedulableSingleSessionContainer: FunctionComponent<SchedulableSingleSess
     return index + result + messageKey
   }
 
+  const canDeleteNotification = (notifications: ScheduleNotification[], index: number) => {
+    // follow-up notifications are always deleteable
+    if (index > 0) return true
```
conundrumgirl, line 221: I think it's easier to read  as `return index> 0 || notifications.length ===1`

src/components/studies/scheduler/NotificationTime.tsx
```diff
@@ -114,6 +114,32 @@ const NotificationTime: React.FunctionComponent<NotificationTimeProps> = ({
...
+    if (
+      isFollowUp &&
+      isMultiday &&
+      prevWindowStartTime !== undefined &&
+      windowStartTime !== undefined &&
+      prevWindowStartTime !== windowStartTime
+    ) {
+      changeMultidayOffset({days: daysForMultidayOffset, time: timeForMultidayOffset})
+    }
+  }, [windowStartTime])
```
conundrumgirl, line 141: I'm a little confused here. If this useEffect only runs when windowStartTime. changes, which it should, since that's what we have in dependency array, why do we need `prevWindowStartTime !== windowStartTime`

hallieswan: I was confused too -- without `prevWindowStartTime !== windowStartTime`, this useEffect is triggered every time the offset is changed, even if windowStartTime is unchanged. 

I think it's because the key for NotificationWindow (parent of NotificationTime) includes the offset. So when the offset changes, a new NotificationWindow instance (containing a new NotificationTime instance) is mounted, so this useEffect is re-run when the new instance is rendered, even though windowStartTime hasn't changed. 

I found [these](https://stackoverflow.com/a/60979832) [two](https://stackoverflow.com/a/54924104) stackoverflow answers to be helpful while troubleshooting

src/components/studies/scheduler/SchedulableSingleSessionContainer.tsx
hallieswan: Agreed! Made the change in the next commit

conundrumgirl (approved):
Nice!

PR #734, opened 2023-09-19
[DHP-1033] Only include “+ Other” response if the type is a single choice string

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/DHP-1033

There were a few bugs and UI/UX issues that I ended up adjusting along with disallowing "other" for a multiple choice question. Basically, there was no differentiation between enabled/disabled for "none of the above" and "all of the above" and that was confusing me. Also there was no explanation for why "none" and "all" weren't allowed for single choice but there was an explanation for "other" only being included for strings. The end result was to add a generic message for "hey, you've run out of special options to add to this question - nothing else applies".

### DIFF
--- a/src/components/surveys/survey-design/question-edit/SelectPhoneBottomMenu.tsx
+++ b/src/components/surveys/survey-design/question-edit/SelectPhoneBottomMenu.tsx
@@ -246,18 +246,18 @@ const SelectPhoneBottomMenu: FunctionComponent<{
     return Array.from(OPTIONS.keys()).filter(optionKey => isOptionSupported(optionKey) && !isOptionAlreadyAdded(optionKey))
   }
 
-  function addMenuItems() {
+  const SpecialOptionsMenuItems: FunctionComponent = () => {
     let options = getOptions()
     return options.length === 0 ? <NoMenuItemOptions /> :
     <>
       {options.map(optionKey => (
-      <StyledMenuItem
-        key={optionKey}
-        disabled={isOptionAlreadyAdded(optionKey)}
-        onClick={() => addGenericResponse(optionKey)}>
-        {OPTIONS.get(optionKey)?.selectDisplayLabel}
-      </StyledMenuItem>
-    ))}
+        <StyledMenuItem
+          key={optionKey}
+          disabled={isOptionAlreadyAdded(optionKey)}
+          onClick={() => addGenericResponse(optionKey)}>
+          {OPTIONS.get(optionKey)?.selectDisplayLabel}
+        </StyledMenuItem>
+      ))}
     </>
   }
 
@@ -306,7 +306,7 @@ const SelectPhoneBottomMenu: FunctionComponent<{
               padding: 0,
             },
           }}>
-          {addMenuItems()}
+          <SpecialOptionsMenuItems />
         </StyledMenu>
       </SideMenu>
     </PhoneBottom>
--- a/src/components/surveys/survey-design/question-edit/SelectPhoneBottomMenu.tsx
+++ b/src/components/surveys/survey-design/question-edit/SelectPhoneBottomMenu.tsx
@@ -1,6 +1,6 @@
 import SurveyUtils from '@components/surveys/SurveyUtils'
 import MoreVertIcon from '@mui/icons-material/MoreVert'
-import {Box, Typography} from '@mui/material'
+import {Box} from '@mui/material'
 
 import AddCircleTwoToneIcon from '@mui/icons-material/AddCircleTwoTone'
 import Button from '@mui/material/Button'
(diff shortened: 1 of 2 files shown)


src/components/surveys/survey-design/question-edit/rhs-subcontrols/Select.tsx
```diff
@@ -148,12 +148,24 @@ const Select: React.FunctionComponent<{
 
     const isSwitchedToSingleSelect = !step.singleChoice && value === 'SINGLE_SELECT'
 
+    const updatedStep = {...step}
+    let choices = isSwitchedToSingleSelect
+    ? step.choices.filter(c => !SurveyUtils.isSpecialSelectChoice(c))
+    : step.choices
+    if (!isSwitchedToSingleSelect) {
+      //if changing to a multiple select -- remove 'other'
```
Erin-Mounts, line 156:
```suggestion
      // if changing to a multiple select -- remove 'other'
```

Erin-Mounts (approved):
one nit, otherwise looks good to me

src/components/surveys/survey-design/question-edit/SelectPhoneBottomMenu.tsx
```diff
@@ -286,18 +294,12 @@ const SelectPhoneBottomMenu: FunctionComponent<{
...
             </StyledMenuItem>
           ))}
-          {step.baseType === 'string' ? (
-            <StyledMenuItem key="OTHER" disabled={isDisabled('OTHER')} onClick={() => addGenericResponse('OTHER')}>
-              + Add "Other"
-            </StyledMenuItem>
-          ) : (
-            <DisabledOtherMenuItem />
-          )}
+          {getOptions().length == 0 ? (<NoMenuItemOptions />) : (<></>) }
```
syoung-smallwisdom, line 302: @hallieswan Is there a way to do this where the UI component takes the filtered available options as an input and returns:

```
if (options.length == 0) {
<DisabledOtherMenuItem />
} else {
options.map(optionKey => (
            <StyledMenuItem
              key={optionKey}
              disabled={isOptionAlreadyAdded(optionKey)}
              onClick={() => addGenericResponse(optionKey)}>
              {OPTIONS.get(optionKey)?.selectDisplayLabel}
            </StyledMenuItem>
          ))
}
```

Or something? I spent way too much time trying to figure that out before giving up. 

Thanks!

src/components/surveys/survey-design/question-edit/rhs-subcontrols/Select.tsx
syoung-smallwisdom: Yeah, this is a weird one. All of Alina's comments don't have a space after the `//` and I thought maybe that was just the standard syntax for code comments?

src/components/surveys/survey-design/question-edit/SelectPhoneBottomMenu.tsx
hallieswan: I think that wrapping the else logic in a [Fragment](https://react.dev/reference/react/Fragment) would work, which would look something like this: 

```TypeScript
{getOptions().length == 0 ? (
  <NoMenuItemOptions />
) : (
    <>
       {getOptions().map(optionKey => (
            <StyledMenuItem
              key={optionKey}
              disabled={isOptionAlreadyAdded(optionKey)}
              onClick={() => addGenericResponse(optionKey)}>
              {OPTIONS.get(optionKey)?.selectDisplayLabel}
            </StyledMenuItem>
          ))}
    </>
)}
```

Side note, this is [a common shortcut](https://react.dev/learn/conditional-rendering#logical-and-operator-) for rendering a component when a condition is true, but nothing otherwise: 

```TypeScript
{getOptions().length == 0 && <NoMenuItemOptions /> }
```

syoung-smallwisdom: Changed to use a function

src/components/surveys/survey-design/question-edit/rhs-subcontrols/Select.tsx
hallieswan: We generally use a space after the `//`, so as far as I know, the missing space isn't a standard syntax.

syoung-smallwisdom: thanks

src/components/surveys/survey-design/question-edit/SelectPhoneBottomMenu.tsx
```diff
@@ -227,15 +232,33 @@ const SelectPhoneBottomMenu: FunctionComponent<{
...
+      {options.map(optionKey => (
+        <StyledMenuItem
+          key={optionKey}
+          disabled={isOptionAlreadyAdded(optionKey)}
+          onClick={() => addGenericResponse(optionKey)}>
+          {OPTIONS.get(optionKey)?.selectDisplayLabel}
+        </StyledMenuItem>
+      ))}
+    </>
   }
```
hallieswan, line 262: Nice! Would it be possible to move `SpecialOptionsMenuItems` out of the parent component (`SelectPhoneBottomMenu`), then pass in props (e.g. `options`), so that `SpecialOptionsMenuItems` is reusable?

syoung-smallwisdom: I'm sure it would be possible, but I couldn't quite figure out how to pass in the properties without it throwing errors. Also the logic for this is tied to the model object specific to the menu items in the ":" menu so it made sense (to me) to only define it internally to the `SelectPhoneBottomMenu`.

PR #736, opened 2023-09-21
DHP-1051: add resolution for tough-cookie

hallieswan:


src/components/surveys/SurveyList.tsx
```diff
@@ -218,6 +218,7 @@ const SurveyList: React.FunctionComponent<{}> = () => {
             <StyledSurveysContainer key="container">
               {surveys?.map((survey, index) => (
                 <Link
+                  component="button"
```
hallieswan, line 221: Test was failing because the `Link` doesn't include an `href` prop, so doesn't have a 'link' role. Instead, render as a button, which is recommended by [MUI for accessibility](https://mui.com/material-ui/react-link/#accessibility) when there is not a meaningful `href`.

package.json
```diff
@@ -100,6 +100,7 @@
     "prettier": "^2.3.2"
   },
   "resolutions": {
-    "minimatch": "^3.1.2"
+    "minimatch": "^3.1.2",
+    "tough-cookie": "^4.1.3"
```
hallieswan, line 104: `tough-cookie` is a dependency of jest, so I confirmed that tests passed before and after updating the `tough-cookie` version.

syoung-smallwisdom: sorry... huh?

syoung-smallwisdom (approved):
looks good? 

Would be nice to have another show-n-tell where we go over how to run tests (and possibly how to include a GitHub action to run them on a pull request.

nickgros (approved):

package.json
hallieswan: Apologies, I should've provided more context -- the repo had a [dependabot alert](https://github.com/Sage-Bionetworks/mtb/security/dependabot/66) that versions of `tough-cookie` <4.1.3 have a moderate severity vulnerability. I ran `yarn why tough-cookie` to see why tough-cookie@4.0.0 was installed, which showed these dependents: react-scripts > jest > @jest/core > jest-environment-jsdom > jsdom > tough-cookie@4.0.0.

The app was set up with [Create React App](https://create-react-app.dev/), which includes `react-scripts` to run tests using [jest](https://jestjs.io/). Since there isn't an updated version of `react-scripts` that uses the patched version of `tough-cookie`, I just set a resolution for `tough-cookie` instead. To confirm that `jest` still ran the tests as expected, I ran the tests with `yarn test` before/after setting the resolution. 

Let me know if you have any questions, hope I answered the right question here. Also, happy to demo running tests and discuss setting up a GitHub action to run tests on PR whenever works best for you.

PR #737, opened 2023-09-22
DHP-1037

hallieswan:
- add bridge login via email/password
- hide behind feature flag
- require re-login upon reload when logging in via email/password
- add hook to get source app info


src/App.tsx
```diff
@@ -51,16 +58,21 @@ function App() {
       <ErrorBoundary FallbackComponent={ErrorFallback} onError={ErrorHandler}>
         {redirect && <Redirect to={redirect}></Redirect>}
         {/*  <React.StrictMode>*/}
-        <FeatureToggleProvider featureToggles={{'SURVEY BUILDER': true}}>
+        <FeatureToggleProvider featureToggles={{'SURVEY BUILDER': true, 'USERNAME PASSWORD LOGIN': false}}>
```
hallieswan, line 61: Set feature toggle for login with username/password to false

src/AuthenticatedApp.tsx
```diff
@@ -59,13 +59,18 @@ const Wrapper: FunctionComponent<RouteComponentProps & {sessionData: UserSession
...
-      sessionStorage.setItem('location', `${location.pathname}${location.search}`)
+    if (sessionData.lastLoginMethod === 'OAUTH_SYNAPSE') {
+      //save location and redirect
+      if (location.pathname !== '/') {
+        sessionStorage.setItem('location', `${location.pathname}${location.search}`)
+      }
+      Utility.redirectToSynapseLogin()
+    } else {
+      sessionUpdateFn({type: 'LOGOUT'})
     }
```
hallieswan, line 73: Continue redirecting to OneSage for users that signed via Synapse OAuth, but log out all other users.

src/App.tsx
```diff
@@ -25,7 +25,14 @@ declare module '@mui/styles/defaultTheme' {
...
 function App() {
-  const {id, isLoading, redirect} = useLogin()
+  const {
+    id,
+    isLoadingLoginWithOauth,
+    isLoadingLoginWithUsernameAndPassword,
+    redirect,
+    submitUsernameAndPassword,
+    errorMessageLoginWithUsernameAndPassword,
+  } = useLogin()
```
hallieswan, line 35: Return additional props from `useLogin` hook, so they can be used in the UsernamePasswordForm via prop drilling. There may be a better way to do this, but leaving that for a future refactor.

src/components/account/AccountLogin.tsx
```diff
@@ -54,24 +60,44 @@ const useStyles = makeStyles(theme => ({
...
-const AccountLogin: FunctionComponent<AccountLoginProps> = ({isArcSignIn}) => {
+const AccountLogin: FunctionComponent<AccountLoginProps> = ({
+  isArcSignIn,
+  submitUsernameAndPassword,
+  isLoadingLoginWithUsernameAndPassword,
+  errorMessageLoginWithUsernameAndPassword,
+}) => {
   //const [alertMsg, setAlertMsg] = useState<{msg: string; type: AlertProps['severity']} | undefined>()
-
+  const sourceApp = useSourceApp()
```
hallieswan, line 70: New hook to get the info about this source app, e.g. long name, short name, etc... We can use this hook when we tackle [DHP-1011](https://sagebionetworks.jira.com/browse/DHP-1011).

src/components/account/AccountLogin.tsx
```diff
@@ -54,24 +60,44 @@ const useStyles = makeStyles(theme => ({
...
-            ? 'Please sign in to ARC using your Synapse account.'
-            : 'Please sign in to Open Bridge using your Synapse account.'}
-        </Box>
+        {createSignInTextBox(`Please sign in to ${sourceApp.longName} with your Synapse account: `)}
         <div className={classes.paper}>
           <Button variant="contained" href={Utility.getRedirectLinkToOneSage()} className={classes.arcSubmitbutton}>
             <SynapseLogo />
           </Button>
         </div>
+        {featureToggles['USERNAME PASSWORD LOGIN'] && (
```
hallieswan, line 91: Use the feature toggle to conditionally display the username password login form.

src/components/account/UsernamePasswordForm.tsx
```diff
@@ -0,0 +1,98 @@
...
+        sx={sx}
+      />
+    </FormControl>
+  )
+}
+
+type UsernamePasswordFormProps = {
+  onSubmit: UseLoginReturn['submitUsernameAndPassword']
+  errorMessage: UseLoginReturn['errorMessageLoginWithUsernameAndPassword']
+  isLoading: UseLoginReturn['isLoadingLoginWithUsernameAndPassword']
```
hallieswan, line 49: Passed down from `useLogin` hook

src/components/widgets/AppTopNav.tsx
```diff
@@ -211,6 +211,11 @@ const AppTopNav: FunctionComponent<AppTopNavProps> = ({
     route => route.name !== 'APP STORE' && route.name !== 'SIGN IN' && (!route.toggle || toggle[route.toggle!] === true)
   )
 
+  const redirectToLoginPage = () => {
+    sessionStorage.setItem('location', `${location.pathname}${location.search}`)
+    window.location.replace(`${window.location.origin}/sign-in`)
+  }
```
hallieswan, line 217: Once we allow sign-in with username/password, users may not have a Synapse account. Redirect user to the Open Bridge sign-in page, rather than directly to OneSage.

src/helpers/AuthContext.tsx
```diff
@@ -44,20 +45,21 @@ function userReducer(state: UserSessionData, action: Action): UserSessionData {
...
+      const newState: UserSessionData = {
         ...state,
         token: action.payload!.token,
         orgMembership: action.payload!.orgMembership,
         appId: action.payload!.appId,
         firstName: action.payload!.firstName,
         lastName: action.payload!.lastName,
-        userName: action.payload!.username,
+        username: action.payload!.username,
         dataGroups: action.payload!.dataGroups,
```
hallieswan, line 56: The field name had a small typo, identified by specifying the type for `newState`.

src/helpers/AuthContext.tsx
```diff
@@ -44,20 +45,21 @@ function userReducer(state: UserSessionData, action: Action): UserSessionData {
...
         roles: action.payload!.roles,
         id: action.payload!.id,
         synapseUserId: action.payload!.synapseUserId,
         isVerified: action.payload!.isVerified,
         demoExternalId: action.payload?.demoExternalId,
+        lastLoginMethod: action.payload?.lastLoginMethod,
```
hallieswan, line 62: `lastLoginMethod` will allow us to determine whether to redirect the user to Synapse or logout the user.

src/useLogin.tsx
```diff
@@ -24,13 +33,52 @@ const attemptLogin = async (code: string): Promise<LoggedInUserData> => {
...
+      const savedLocation = sessionStorage.getItem('location')
+      if (savedLocation) {
+        sessionStorage.removeItem('location')
+        setRedirect(savedLocation)
+      } else {
+        setRedirect('/studies')
+      }
+    },
+    [sessionUpdateFn]
+  )
```
hallieswan, line 80: Abstracted from the `useEffect` that watches for the OAuth code, so it can be reused when logging with username/password.

src/App.tsx
syoung-smallwisdom: Can you add a TODO comment (signed and dated) to refactor to a keyed mapping rather than the list of implied "only one" bools? Thanks!

src/App.tsx
syoung-smallwisdom: Is this how the feature is hidden? Do we have a variable in the model for whether or not this is staging? Would be nice to show it on staging but hide on production.

src/AuthenticatedApp.tsx
syoung-smallwisdom: Can you add a TODO comment (signed and dated) with the JIRA ticket for allowing reauth: https://sagebionetworks.jira.com/browse/DHP-1057

src/components/account/AccountLogin.tsx
syoung-smallwisdom: Can you and a TODO comment (signed and dated) with the JIRA ticket? Thanks!

src/types/sourceAppConfig.ts
```diff
@@ -0,0 +1,32 @@
+import {useUserSessionDataState} from '@helpers/AuthContext'
+import {ARC_APP_ID, INV_ARC_ID, MTB_APP_ID, OPEN_BRIDGE_APP_ID} from './constants'
+
+type SourceAppConfig = {
+  longName: string
+  shortName: string
+}
+
+export const sourceAppConfig: Record<string, SourceAppConfig> = {
```
syoung-smallwisdom, line 9: Nice!

syoung-smallwisdom (approved):
Lots of TODO comment asks and one question about the feature toggling.

Otherwise, looks good.

src/SignInPage.tsx
```diff
@@ -46,10 +47,18 @@ const useStyles = makeStyles(theme => ({
 }))
 
 type SignInPageProps = {
+  isLoadingLoginWithUsernameAndPassword: UseLoginReturn['isLoadingLoginWithUsernameAndPassword']
+  submitUsernameAndPassword: UseLoginReturn['submitUsernameAndPassword']
+  errorMessageLoginWithUsernameAndPassword: UseLoginReturn['errorMessageLoginWithUsernameAndPassword']
```
jay-hodgson, line 52: Is there a cleaner way to inherit these fields from UseLoginReturn?

src/App.tsx
hallieswan: Yes, this is how features are hidden right now. There's an `isDevelopment` method that could be used to show locally / on staging, but hide on prod -- I'll update in the next commit. I also opened [a ticket](https://sagebionetworks.jira.com/browse/DHP-1065) to track setting up a button to allow the user to toggle the features, if desired.

src/App.tsx
hallieswan: Jay noted a similar issue, so I've refactored to pass down an object of the username/password variables, rather than passing down each variable separately. Happy to add a TODO if that doesn't cover the issue noted here!

src/SignInPage.tsx
hallieswan: I've refactored to pass down an object with these props, rather than passing each down separately. Happy to update if there's a better approach!

src/App.tsx
```diff
@@ -58,7 +52,7 @@ function App() {
       <ErrorBoundary FallbackComponent={ErrorFallback} onError={ErrorHandler}>
         {redirect && <Redirect to={redirect}></Redirect>}
         {/*  <React.StrictMode>*/}
-        <FeatureToggleProvider featureToggles={{'SURVEY BUILDER': true, 'USERNAME PASSWORD LOGIN': false}}>
```
hallieswan: Show username/password login locally and on staging, but no on production.

src/components/account/AccountLogin.tsx
hallieswan: Added TODO in `types/sourceAppConfig.ts`

src/index.tsx
```diff
@@ -6,15 +6,15 @@ import {createRoot} from 'react-dom/client'
...
+export function isDevelopment(): boolean {
   return (
     window.location.hostname === 'localhost' ||
     // [::1] is the IPv6 localhost address.
     window.location.hostname === '[::1]' ||
     // or we are on staging
     window.location.hostname.includes('staging') ||
     // 127.0.0.0/8 are considered localhost for IPv4.
-    window.location.hostname.match(/^127(?:\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}$/)
+    window.location.hostname.match(/^127(?:\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}$/) !== null
```
hallieswan, line 17: Ensure that a boolean is returned

src/useLogin.tsx
```diff
@@ -4,13 +4,17 @@ import Utility from './helpers/utility'
 import UserService from './services/user.service'
 import {ExtendedError, LoggedInUserData, LoginMethod} from './types/types'
 
+export type UsernameAndPasswordLogin = {
+  isLoading: boolean
+  errorMessage: string | undefined
+  onSubmit: (username: string, password: string) => void
+}
```
hallieswan, line 11: Simplify prop drilling by grouping the variables for the UsernameAndPasswordForm in the same object

src/AuthenticatedApp.tsx
```diff
@@ -59,13 +59,20 @@ const Wrapper: FunctionComponent<RouteComponentProps & {sessionData: UserSession
 
 const AuthenticatedApp: FunctionComponent<RouteComponentProps> = ({location, match}) => {
   const sessionData = useUserSessionDataState()
+  const sessionUpdateFn = useUserSessionDataDispatch()
 
   if (!sessionData.token) {
```
jay-hodgson, line 64: Is sessionData.token ever set?  Why is session storage used at all (couldn't it just rely on the in-memory context, and send the user to login if not set)?
I'm not recommending a rewrite as part of this issue/PR, but these are questions that the Bridge team may want to follow-up on in the future.

jay-hodgson (approved):

src/AuthenticatedApp.tsx
hallieswan: `sessionData.token` is only set in React context -- when sessionData is saved to sessionStorage in [`AuthContext`](https://github.com/hallieswan/mtb/blob/91bc7f54d417629cca16b393e4205bad061bde1f/src/helpers/AuthContext.tsx#L65), the token is stored as `undefined`. I agree that we could rely on the in-memory context for the auth token and send the user to login if not set. sessionStorage is used by two methods (`isInAdminRole`, `isPathAllowed`) to get the user's `AdminRoles`, but we could refactor. I opened [a ticket](https://sagebionetworks.jira.com/browse/DHP-1083) to track the work for future follow-up.

PR #744, opened 2023-10-02
DHP-1068 Hide passive recorders by default

syoung-smallwisdom:
Hide passive recorders by default.

Turns out that these were only hidden for Arc.


Erin-Mounts (approved):
looks good?

hallieswan (approved):
Looks good to me!

PR #747, opened 2023-10-03
DHP-1053 Do not add an assessment to the list of assessments

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/DHP-1053

When copying, a copy of the assessment was being added with the identifier changed to "..." - the `guid`, which is the identifier used by Bridge, was not replaced by the "placeholder" and this copy was never removed.

Opened https://sagebionetworks.jira.com/browse/DHP-1093 to track improving the UI/UX for copying surveys


Erin-Mounts (approved):
:+1:

PR #748, opened 2023-10-04
Add surveys to schedule

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/DHP-1091

## Terminology Reminder
Within the Researcher UI, the following are equivalent:
- "Survey" means a local assessment where all the calls need to use the `/v1/assessment` or  `/v1/assessments` endpoints.
- "Assessment" means a shared assessment where all the calls need to use either `/v1/sharedassessment` or  `/v1/sharedassessments` endpoints.

Since the model object returned by both of these services is the same JSON, the website needs to track whether or not the origin is "shared" or "local". Thanks to @hallieswan for explaining how to subclass the server `Assessment` to add the required field for `isLocal` so that subsequent calls to get information about the assessment can point at the correct services.


Erin-Mounts (approved):
The parts I can follow look fine to me....

PR #750, opened 2023-10-06
DHP-1089 Fix "Rename" survey to actually allow renaming

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/DHP-1089 - The issue that I was originally looking at was that if you are on the survey list view and you select "Rename" from the menu, it doesn't work. Instead it was not hiding the menu and would navigate to the detail view for the survey.

I tried to match what it was doing on the Study list view (the code for "Survey Builder" list was copy/pasted from that code file), discovered a ton of warnings in tests and other discrepancies, and ended up down a bit of a renaming rabbit hole.

That said, the main "fixes" were:
- call through to services with the correct property names ("assessment", not "survey")
- don't navigate to the detail if the "Title" field is highlighted (ie. textfield rather than label)

Note: The UI/UX for a "rename" of a survey or study is a bit weird. There's no explicit blocking pop-up like there is on Google drive so you can sit there clicking away all day and you are blocked from selecting that study or survey in order to view the detail until:

1.  You have tapped into the the text field to focus that input field
2. And then you enter (using your keyboard) either "esc" to cancel editing the title or "return" or "tab" to accept the changes.

I believe this is "by design" and not (strictly speaking) a bug.

### DIFF
--- a/src/components/assessments/AssessmentLibrary.tsx
+++ b/src/components/assessments/AssessmentLibrary.tsx
@@ -47,7 +47,7 @@ const AssessmentLibrary: FunctionComponent<AssessmentLibraryProps> = ({match}: A
       {data && (
         <AssessmentLibraryWrapper
           assessments={data.assessments}
-          assessmentsType="OTHER"
+          assessmentsType='SHARED'
           viewMode={viewMode}
           onChangeViewMode={setViewMode}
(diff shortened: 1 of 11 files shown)


Erin-Mounts (approved):
looks good to me

PR #751, opened 2023-10-09
DHP-1079

hallieswan:
https://sagebionetworks.jira.com/browse/DHP-1079
- use radio group for minimum and maximum year controls
- add tests for year control
- specify a key for the RHS control, so children are re-rendered when switching between steps of the same question type

https://github.com/BridgeDigitalHealth/open-bridge-web/assets/26949006/218e000a-8eaa-4b3b-ae1e-046c366e7db1

### DIFF
--- a/src/components/surveys/survey-design/question-edit/QuestionEditRhs.tsx
+++ b/src/components/surveys/survey-design/question-edit/QuestionEditRhs.tsx
@@ -134,6 +134,9 @@ const QuestionEditRhs: FunctionComponent<QuestionEditProps> = ({step, onChange, 
               onChange: onChange,
               q_type: getQuestionId(step),
             }}
+            // TODO: Year and Time are partially controlled, so use a key to
+            // re-render the entire component when the step is changed.
+            // Consider making these components fully controlled instead.
             key={step.identifier}></Factory>
         </Box>
       </div>
--- a/src/components/surveys/survey-design/question-edit/QuestionEditRhs.tsx
+++ b/src/components/surveys/survey-design/question-edit/QuestionEditRhs.tsx
@@ -133,7 +133,8 @@ const QuestionEditRhs: FunctionComponent<QuestionEditProps> = ({step, onChange, 
               step: {...step},
               onChange: onChange,
               q_type: getQuestionId(step),
-            }}></Factory>
+            }}
+            key={step.identifier}></Factory>
         </Box>
       </div>
       {children}
--- a/src/components/surveys/survey-design/question-edit/rhs-subcontrols/Year.test.tsx
+++ b/src/components/surveys/survey-design/question-edit/rhs-subcontrols/Year.test.tsx
@@ -1,18 +1,18 @@
-import {screen, waitFor, within} from '@testing-library/react'
+import {fireEvent, render, screen, waitFor, within} from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import {FormatOptionsYear, YearQuestion} from '@typedefs/surveys'
-import {renderSurveyQuestionComponent} from '__test_utils/utils'
+import {createWrapper, renderSurveyQuestionComponent} from '__test_utils/utils'
 import Year, {ErrorMessages} from './Year'
 import {DEFAULT_MAX_YEAR, DEFAULT_MIN_YEAR} from './YearRadioGroup'
+
+const CURRENT_YEAR = new Date().getFullYear()
 
 //render the component
 const renderComponent = (step: YearQuestion) => {
   return renderSurveyQuestionComponent<YearQuestion>({step, Component: Year})
 }
 
-// render component and return controls
-function setUp(step: YearQuestion) {
-  const {user, element: component} = renderComponent(step)
-
+const getYearElements = () => {
   const minYearFormatGroup = screen.getByRole('radiogroup', {name: 'Minimum Year'})
   const minYear = {
     buttons: {
@@ -32,8 +32,26 @@ function setUp(step: YearQuestion) {
     },
     value: document.getElementById('maxYearValue'),
   }
+  return {minYear, maxYear}
+}
+
+// render component and return controls
+function setUp(step: YearQuestion) {
+  const {user, element: component} = renderComponent(step)
+
+  const {minYear, maxYear} = getYearElements()
 
   return {component, user, minYear, maxYear}
+}
+
+const setUpWithChangeMock = (step: YearQuestion) => {
+  const onChangeMock = jest.fn()
+  const user = userEvent.setup()
+  const component = render(<Year step={step} onChange={onChangeMock} />, {
+    wrapper: createWrapper(),
+  })
+  const {minYear, maxYear} = getYearElements()
+  return {user, component, minYear, maxYear, onChangeMock}
 }
 
 // default props
@@ -179,10 +197,9 @@ describe('Year', () => {
   })
 
   test('error is shown when allowPast is false and maximumYear is less than current year', async () => {
-    const currentYear = new Date().getFullYear()
-    const errorYear = currentYear - 1
+    const errorYear = CURRENT_YEAR - 1
 
-    const {user, maxYear} = setUp(createStepProps({allowPast: false, maximumYear: currentYear}))
+    const {user, maxYear} = setUp(createStepProps({allowPast: false, maximumYear: CURRENT_YEAR}))
     expect(screen.queryByRole('alert')).not.toBeInTheDocument()
 
     await waitFor(async () => {
@@ -196,10 +213,9 @@ describe('Year', () => {
   })
 
   test('error is shown when allowFuture is false and minimumYear is greater than current year', async () => {
-    const currentYear = new Date().getFullYear()
-    const errorYear = currentYear + 1
+    const errorYear = CURRENT_YEAR + 1
 
-    const {user, minYear} = setUp(createStepProps({allowFuture: false, minimumYear: currentYear}))
+    const {user, minYear} = setUp(createStepProps({allowFuture: false, minimumYear: CURRENT_YEAR}))
     expect(screen.queryByRole('alert')).not.toBeInTheDocument()
 
     await waitFor(async () => {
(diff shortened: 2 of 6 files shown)


src/components/surveys/survey-design/question-edit/QuestionEditRhs.tsx
```diff
@@ -133,7 +133,8 @@ const QuestionEditRhs: FunctionComponent<QuestionEditProps> = ({step, onChange,
               step: {...step},
               onChange: onChange,
               q_type: getQuestionId(step),
-            }}></Factory>
+            }}
+            key={step.identifier}></Factory>
```
hallieswan, line 137: Specify a key, so that child components are re-rendered when switching between steps of the same question type.

Before -- the inputs match the last question, rather than retaining the original inputs for that step.

https://github.com/BridgeDigitalHealth/open-bridge-web/assets/26949006/8dba45ac-6100-4e74-8acb-73e7577bdf77

After -- unique inputs per step are retained when switching between steps.

https://github.com/BridgeDigitalHealth/open-bridge-web/assets/26949006/c59df58d-c84e-48d3-971a-db0800fa8413

src/components/surveys/survey-design/question-edit/rhs-subcontrols/Year.tsx
```diff
@@ -110,54 +66,56 @@ const Year: React.FunctionComponent<{
     if (range.max && range.min && range.max < range.min) {
       return 'RANGE'
     }
-    if (gtCurrent && !allowFuture) {
+    if (gtCurrent && allowFuture !== undefined && !allowFuture) {
```
hallieswan, line 69: Handle case where `allowFuture` is undefined when a maximum year value is set for the maximum year radio group.

src/components/surveys/survey-design/question-edit/rhs-subcontrols/Year.tsx
```diff
@@ -1,78 +1,36 @@
...
-      htmlFor={CONFIG[type].labelId}
-      sx={{mt: theme.spacing(1.5)}}
-      control={<Checkbox checked={value !== false} onChange={e => onChange(e.target.checked)} />}
-      label={<Typography sx={{fontWeight: '14px'}}>{CONFIG[type].label}</Typography>}
-    />
-  )
+const yearFormatToAllowValueMap: Record<YearFormatType, boolean | undefined> = {
+  SET: undefined,
+  ANY: true,
+  CURRENT: false,
```
hallieswan, line 14: Create a mapping of the selected radio group value to the "allow" value, e.g. `allowPast`

src/components/surveys/survey-design/question-edit/rhs-subcontrols/YearRadioGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+type YearRadioGroupProps = {
+  type: 'min' | 'max'
+  yearFormat: YearFormatType
+  yearValue: number | undefined
+  hasError: boolean
+  onChange: (yearFormat: YearFormatType, yearValue: number | undefined) => void
+}
+
+const YearRadioGroup: React.FunctionComponent<YearRadioGroupProps> = ({
```
hallieswan, line 20: Presentational component that calls `onChange` when a new year format is selected or when a year value is updated.

src/components/surveys/survey-design/question-edit/rhs-subcontrols/Year.tsx
```diff
@@ -1,78 +1,36 @@
...
-          sx={{width: '80px'}}
-          id={CONFIG[type].labelId}
-          error={hasError}
-          value={value ?? ''}
-          type="number"
-          //@ts-ignore
-          onChange={(e: ChangeEvent<any>) => {
-            onChange(parseInt(e.target.value))
-          }}
-        />
```
hallieswan, line 69: Moved into `YearRadioGroup` component

src/components/surveys/survey-design/question-edit/rhs-subcontrols/YearRadioGroup.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+const capitalizeFirstLetter = (val: string) => {
+  return `${val.charAt(0).toUpperCase()}${val.substring(1)}`
+}
+
+const yearFormatToAllowValueMap: Record<YearFormatType, boolean | undefined> = {
+  SET: undefined,
+  ANY: true,
+  CURRENT: false,
+}
```
hallieswan, line 19: Mapping between the allowed value (e.g. `allowPast`) and the selected year format.

src/components/surveys/survey-design/question-edit/rhs-subcontrols/YearRadioGroup.tsx
```diff
@@ -0,0 +1,112 @@
...
+    min: {
+      tense: 'past',
+      yearLabel: 'Minimum Year',
+    },
+    max: {
+      tense: 'future',
+      yearLabel: 'Maximum Year',
+    },
+  }
+  const yearFormat: YearFormatType = allowValueToYearFormat(allowValue)
```
hallieswan, line 63: Map the allowValue to the radio group button that should be selected

src/components/surveys/survey-design/question-edit/rhs-subcontrols/YearRadioGroup.tsx
```diff
@@ -0,0 +1,112 @@
...
+    <>
+      <FormControl fullWidth sx={{marginTop: 2}} error={hasError}>
+        <SimpleTextLabel id={id} sx={{marginBottom: 1}}>{`${CONFIG[type].yearLabel}`}</SimpleTextLabel>
+        <RadioGroup
+          aria-labelledby={id}
+          id={`${type}YearFormat`}
+          value={yearFormat}
+          onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+            const newYearFormat = e.target.value as YearFormatType
+            onChange(yearFormatToAllowValue(newYearFormat), setYearValue(newYearFormat, yearValue || defaultYearValue))
```
hallieswan, line 96: Use `onChange` callback to send the updated allowValue and yearValue to the parent component.

src/components/surveys/survey-design/question-edit/QuestionEditRhs.tsx
nickgros: Seems like the real issue is probably that the `Year` component doesn't update when it gets a new `step` prop? It looks like `Year` is neither fully controlled or fully uncontrolled--you probably want to make it fully controlled. This is an ok stopgap if that isn't a simple change (or if some of the other `Factory` components have the same bug)

src/components/surveys/survey-design/question-edit/rhs-subcontrols/Year.test.tsx
```diff
@@ -32,87 +52,201 @@ const step: YearQuestion = {
...
+    await waitFor(async () => {
+      await user.dblClick(minYear.value!)
+      await user.keyboard(errorYear.toString())
+    })
+    expect(minYear.value).toHaveValue(errorYear)
+    const alert = screen.queryByRole('alert')
+    expect(alert).toBeInTheDocument()
+    expect(alert).toHaveTextContent(ErrorMessages['RANGE'])
+  })
+})
```
nickgros, line 252: Nice set of tests. You could consider adding a test for the stale data bugfix if you implement a fix for the `Year` component.

src/components/surveys/survey-design/question-edit/rhs-subcontrols/Year.tsx
```diff
@@ -110,54 +40,56 @@ const Year: React.FunctionComponent<{
...
+      }
+      // only call onChange if the formatOptions have actually changed
+      if (!_.isEqual(formatOptions, prevFormatOptions)) {
+        setPrevFormatOptions(formatOptions)
+        const inputItem = {...step.inputItem, formatOptions: formatOptions}
+        onChange({...step, inputItem})
+      }
     }
-  }, [range, allowPast, allowFuture, onUpdateFormat, step.inputItem])
+  }, [range, allowPast, allowFuture, prevFormatOptions, error, onChange, step])
```
nickgros, line 70: If you make this fully controlled, you could probably call `validate` and `onChange` in the `YearRadioGroup`'s `onChange` prop and then completely remove this effect (and not have to worry about an entire class of bugs)

src/components/surveys/survey-design/question-edit/QuestionEditRhs.tsx
syoung-smallwisdom: If you are using the step.identifier as the key, then what happens when you change the step identifier?

syoung-smallwisdom (approved):
looks good?

- Functionally it works as expected
- I have a question about using the step identifier as the key

Also, is it possible to add checking the JSON and/or model when performing unit tests? In Android, UI tests allow for this, but on iOS you aren't given access to this and have to separate UI tests, unit tests, and integration tests. I ask b/c the logic for setting the JSON is kinda weird. Thanks!

nickgros (approved):
Think we decided that the `key` is appropriate, and removing it would probably be an unnecessarily large lift. Recommend adding a TODO to note that Year/Time components are not "fully controlled" as one might expect

src/components/surveys/survey-design/question-edit/QuestionEditRhs.tsx
hallieswan: @nickgros: agreed! For now, I've added TODO comments to make Year and Time fully controlled components.

@syoung-smallwisdom: when the step identifier is updated in `QuestionEditRhs`, the `survey` state (which includes the current `step`) is updated in `SurveyDesign` via a callback. Since the state was updated, `SurveyDesign` is re-rendered and the updated `step` is passed as a prop to `QuestionEditRhs` and then to `Factory`. Since `step` has a new identifier and thus a new key, the state within the `Factory` component is [reset](https://react.dev/reference/react/useState#resetting-state-with-a-key). 

For `Year`, the state will be initialized using the current values of `step`, which will match the last valid configuration for the Year. 

Last state was valid:

https://github.com/BridgeDigitalHealth/open-bridge-web/assets/26949006/863aa7e8-de32-4b69-8bfd-67946752257f

Last state was invalid:

https://github.com/BridgeDigitalHealth/open-bridge-web/assets/26949006/49522563-5a00-44ce-82d5-fb396af25642

syoung-smallwisdom:
Thanks for adding the tests to mock the change and check the model!

src/components/surveys/survey-design/question-edit/QuestionEditRhs.tsx
syoung-smallwisdom: Thanks!

PR #759, opened 2023-10-13
DHP-1114 Block changes to branching logic if read-only

syoung-smallwisdom:
This uses the same blocking UI/UX as in the survey where drop-down and radio button elements are unchanged but cannot be edited. Turns out that b/c the Assessment and the AssessmentConfig are two different service calls, they have to be re-fetched (from cache) because the branching logic has multiple "routing" urls.

Screenshot 2023-10-13 at 3 19 56 PM


Erin-Mounts (approved):
looks good to me

PR #762, opened 2023-10-20
Revert "Move ReadOnlyBanner to main content"

syoung-smallwisdom:
This reverts commit 19e872f75a9969cf8906ed776285b7510db20e5e.

No idea why but apparently moving the "read-only" banner to across the top breaks text input field editing. For now, revert the change and just live with the undiscoverable banner. ¯\_(ツ)_/¯


Erin-Mounts (approved):
okay

PR #763, opened 2023-10-20
move read only banner and add explicit "label" field

syoung-smallwisdom:
- Figured out how to add the read-only banner without breaking everything
- I also fixed the title page (overview) text field so the title will wrap to multiple lines
- Moved the "label" shown to participants to the "settings" page so that it's explicitly called out
- Fixed a bug where there was no feedback to the user that they made changes on the settings page

Moving the "label" to the "settings" does two things:
1. It's explicit 
2. Improves performance b/c the assessment isn't being updated on the server with every key press


Erin-Mounts (approved):
looks good

PR #764, opened 2023-10-23
DHP-1045: add ability to request password reset

hallieswan:
https://sagebionetworks.jira.com/browse/DHP-1045

https://github.com/BridgeDigitalHealth/open-bridge-web/assets/26949006/0931ac7e-2517-4e19-9c84-e551c72ff838

### DIFF
--- a/src/SignInPage.tsx
+++ b/src/SignInPage.tsx
@@ -1,5 +1,5 @@
 import {ResetPassword} from '@components/account/ResetPassword'
-import useFeatureToggles, {FeatureToggles} from '@helpers/FeatureToggle'
+import useFeatureToggles, {FeatureToggles, features} from '@helpers/FeatureToggle'
 import {Box} from '@mui/material'
 import makeStyles from '@mui/styles/makeStyles'
 import constants from '@typedefs/constants'
@@ -86,9 +86,8 @@ const SignInPage: React.FunctionComponent<SignInPageProps> = ({isARCApp, usernam
             usernameAndPasswordLogin={usernameAndPasswordLogin}
           />
         )}
-        {featureToggles['USERNAME PASSWORD LOGIN'] && location.pathname === constants.publicPaths.RESET_PASSWORD && (
-          <ResetPassword />
-        )}
+        {featureToggles[features.USERNAME_PASSWORD_LOGIN] &&
+          location.pathname === constants.publicPaths.RESET_PASSWORD && <ResetPassword />}
       </Box>
     </Box>
   )
--- a/src/SignInPage.tsx
+++ b/src/SignInPage.tsx
@@ -1,7 +1,11 @@
+import {ResetPassword} from '@components/account/ResetPassword'
+import useFeatureToggles, {FeatureToggles} from '@helpers/FeatureToggle'
 import {Box} from '@mui/material'
 import makeStyles from '@mui/styles/makeStyles'
+import constants from '@typedefs/constants'
 import clsx from 'clsx'
 import React from 'react'
+import {useLocation} from 'react-router-dom'
 import {UseLoginReturn} from 'useLogin'
 import ArcLogo from './assets/logo_arc_main.svg'
 import MtbFinalLogo from './assets/logo_open_bridge_large.svg'
@@ -53,6 +57,9 @@ type SignInPageProps = {
 
 const SignInPage: React.FunctionComponent<SignInPageProps> = ({isARCApp, usernameAndPasswordLogin}) => {
   const classes = useStyles()
+  const location = useLocation()
+  const featureToggles = useFeatureToggles<FeatureToggles>()
+
   return (
     <Box className={clsx(classes.container, !isARCApp && classes.mtbContainer)}>
       {isARCApp && (
@@ -72,11 +79,16 @@ const SignInPage: React.FunctionComponent<SignInPageProps> = ({isARCApp, usernam
         </Box>
       )}
       <Box className={classes.rightContainer} sx={{width: isARCApp ? '50%' : '100%'}}>
-        <AccountLogin
-          callbackFn={() => {}}
-          isArcSignIn={isARCApp}
-          usernameAndPasswordLogin={usernameAndPasswordLogin}
-        />
+        {(location.pathname === constants.publicPaths.SIGN_IN || location.pathname === '/') && (
+          <AccountLogin
+            callbackFn={() => {}}
+            isArcSignIn={isARCApp}
+            usernameAndPasswordLogin={usernameAndPasswordLogin}
+          />
+        )}
+        {featureToggles['USERNAME PASSWORD LOGIN'] && location.pathname === constants.publicPaths.RESET_PASSWORD && (
+          <ResetPassword />
+        )}
       </Box>
     </Box>
   )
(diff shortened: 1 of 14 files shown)


src/components/account/UsernamePasswordForm.tsx
```diff
@@ -59,40 +61,43 @@ const UsernamePasswordForm: FunctionComponent<UsernamePasswordFormProps> = ({onS
...
-        }}>
-        {isLoading ? 'Logging you in...' : 'Sign In'}
-      </Button>
-      {errorMessage && (
-        <Alert variant="outlined" color="error" sx={{mt: 2, alignItems: 'center'}}>
-          {errorMessage}
-        </Alert>
-      )}
-    </form>
+    <>
```
hallieswan, line 64: The diff isn't very useful here -- the only changes are 1) wrapping the form in an empty tag, so that 2) the "Forgot password?" nav link can be added below the form.

src/routes_public.ts
```diff
@@ -33,10 +34,16 @@ const routes = [
...
   {
-    path: '/sign-in',
+    path: constants.publicPaths.SIGN_IN,
     name: 'SIGN IN',
     Component: SignInPage,
     isRhs: true,
   },
+  {
+    path: constants.publicPaths.RESET_PASSWORD,
+    name: '',
```
hallieswan, line 44: `name` needs to be an empty string, so that the reset password route doesn't appear in the navbar.

src/components/account/ResetPassword.tsx
```diff
@@ -0,0 +1,67 @@
...
+      })
+  }
+
+  const handleCloseMessage = () => {
+    setShowMessage(false)
+  }
+
+  return (
+    <Container component="main" maxWidth="xs" sx={{display: 'flex', flexDirection: 'column', alignItems: 'center'}}>
+      <Typography variant="h3" mb={2} sx={{fontFamily: poppinsFont}}>
```
hallieswan, line 36: Match the font used on the sign in page

src/SignInPage.tsx
```diff
@@ -72,11 +79,16 @@ const SignInPage: React.FunctionComponent<SignInPageProps> = ({isARCApp, usernam
...
-          usernameAndPasswordLogin={usernameAndPasswordLogin}
-        />
+        {(location.pathname === constants.publicPaths.SIGN_IN || location.pathname === '/') && (
+          <AccountLogin
+            callbackFn={() => {}}
+            isArcSignIn={isARCApp}
+            usernameAndPasswordLogin={usernameAndPasswordLogin}
+          />
+        )}
+        {featureToggles['USERNAME PASSWORD LOGIN'] && location.pathname === constants.publicPaths.RESET_PASSWORD && (
```
nickgros, line 89: Not very familiar with FeatureToggles in this app, but `'USERNAME PASSWORD LOGIN'` seems like a good candidate for a constant

src/components/account/ResetPassword.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return {user, spyOnRequestResetPassword, element, emailField, submitButton}
+}
+
+afterEach(() => {
+  jest.restoreAllMocks()
+  cleanup()
+})
+
+describe.only('ResetPassword', () => {
```
nickgros, line 24: Just curious, why `describe.only`? Guessing maybe it's a convention to enforce only one top-level `describe` per file?

src/components/account/ResetPassword.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  })
+
+  test('should show success banner when password reset request is sent', async () => {
+    const {user, emailField, submitButton, spyOnRequestResetPassword} = setUp()
+    const testEmail = 'test@fake.com'
+
+    await waitFor(async () => {
+      await user.type(emailField, testEmail)
+      await user.click(submitButton)
+    })
```
nickgros, line 39: This `waitFor` is probably not necessary since it will just re-run if either of these actions throw an error. If either of these actions _do_ throw, then you should probably `await findBy...` the element that hasn't appeared yet, instead of wrapping the actions.

src/components/account/ResetPassword.tsx
```diff
@@ -0,0 +1,67 @@
...
+    await UserService.requestResetPassword(email)
+      .then(result => {
+        setMessage(result.data.message)
+      })
+      .catch((err: Error) => {
+        setMessage(err.message)
+      })
+      .finally(() => {
+        setShowMessage(true)
+      })
```
nickgros, line 27: Should not use both `await` and `.then/.catch/.finally` callbacks.

Either remove `await` or keep it and wrap in `try { } catch(e) {} finally {}`

nickgros (approved):
Looks good, had a few suggestions

syoung-smallwisdom (approved):
looks good

src/components/account/ResetPassword.test.tsx
hallieswan: Oh good catch! This is a holdover from writing the tests, not a convention of the repo -- I'll remove it!

src/components/account/ResetPassword.test.tsx
hallieswan: Good call! I moved the actions outside of the `waitFor`. Since the `emailField` was already returned via a `getBy` query in the `setUp`, I then `waitFor` the `emailField` to be updated with the test email address -- which I think adds up to a `findBy`, since `findBy` is a combination of `getBy*` queries and `waitFor`.

src/helpers/FeatureToggle.tsx
```diff
@@ -1,6 +1,12 @@
 import React, {useEffect, useState} from 'react'
 
-export type ToggleKey = 'SURVEY BUILDER' | 'USERNAME PASSWORD LOGIN' | 'SHARED'
+export const features = {
+  SURVEY_BUILDER: 'SURVEY BUILDER',
+  USERNAME_PASSWORD_LOGIN: 'USERNAME PASSWORD LOGIN',
+  SHARED: 'SHARED',
+} as const
+
+export type ToggleKey = typeof features[keyof typeof features]
```
hallieswan, line 9: Move features into an object, so that feature toggles can be set using a constant.

src/components/account/ResetPassword.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const {user, emailField, submitButton, spyOnRequestResetPassword} = setUp()
+    const testEmail = 'test@fake.com'
+
+    user.type(emailField, testEmail)
+    await waitFor(() => {
+      expect(emailField).toHaveValue(testEmail)
+    })
+
+    expect(submitButton).not.toBeDisabled()
+    user.click(submitButton)
```
nickgros, line 42: Should still `await` the `user` actions

```suggestion
    await user.type(emailField, testEmail)
    await waitFor(() => {
      expect(emailField).toHaveValue(testEmail)
    })

    expect(submitButton).not.toBeDisabled()
    await user.click(submitButton)
```

nickgros (approved):

src/components/account/ResetPassword.test.tsx
hallieswan: Calling `await` before `click` or `type` causes some `act` warnings to appear -- this seems to be related to testing-library packages using different version of @testing-library/dom, which have different configurations. I've opened a [ticket](https://sagebionetworks.jira.com/browse/DHP-1123) to track the issue.

PR #769, opened 2023-10-25
DHP-1128 Fix question choice values and UI key mapping

syoung-smallwisdom:
- integer-type choices should be numbers and string values should be strings
- when changing the value, need to update survey rules and keep UI key map consistent

https://sagebionetworks.jira.com/browse/DHP-1128

### DIFF
--- a/src/components/surveys/survey-design/SurveyDesign.tsx
+++ b/src/components/surveys/survey-design/SurveyDesign.tsx
@@ -23,11 +23,12 @@ import {SURVEY_ICONS} from '../widgets/SurveyIcon'
 import IntroInfo from './IntroInfo'
 import AddQuestionMenu from './left-panel/AddQuestionMenu'
 import LeftPanel from './left-panel/LeftPanel'
-import QUESTIONS, {QuestionTypeKey} from './left-panel/QuestionConfigs'
+import QUESTIONS, {QuestionTypeKey, getQuestionId} from './left-panel/QuestionConfigs'
 import QuestionEditPhone from './question-edit/QuestionEditPhone'
 import QuestionEditRhs from './question-edit/QuestionEditRhs'
 import QuestionEditToolbar from './question-edit/QuestionEditToolbar'
 import ReadOnlyBanner from '@components/widgets/ReadOnlyBanner'
+import { ConnectWithoutContactOutlined } from '@mui/icons-material'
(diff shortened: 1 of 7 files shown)


Erin-Mounts (approved):
looks good to me

src/components/surveys/survey-design/SurveyDesign.tsx
```diff
@@ -23,11 +23,12 @@ import {SURVEY_ICONS} from '../widgets/SurveyIcon'
 import IntroInfo from './IntroInfo'
 import AddQuestionMenu from './left-panel/AddQuestionMenu'
 import LeftPanel from './left-panel/LeftPanel'
-import QUESTIONS, {QuestionTypeKey} from './left-panel/QuestionConfigs'
+import QUESTIONS, {QuestionTypeKey, getQuestionId} from './left-panel/QuestionConfigs'
 import QuestionEditPhone from './question-edit/QuestionEditPhone'
 import QuestionEditRhs from './question-edit/QuestionEditRhs'
 import QuestionEditToolbar from './question-edit/QuestionEditToolbar'
 import ReadOnlyBanner from '@components/widgets/ReadOnlyBanner'
+import { ConnectWithoutContactOutlined } from '@mui/icons-material'
```
hallieswan, line 31: Just curious, are the two new imports (`getQuestionId`, `ConnectWithoutContactOutlined`) needed? I didn't seem them used in the changes, but I might've missed them!

src/components/surveys/survey-design/question-edit/rhs-subcontrols/Select.tsx
```diff
@@ -172,30 +217,49 @@ const Select: React.FunctionComponent<{
...
+        choice: choice, 
+        index: ii,
+        newValue:  generateValue(choice, isSwitchToInteger ? ii : undefined, value)
+      }))
+      
+      // apply the changes to both choices and survey rules
+      const choices = getNewChoices(changes)
+      const surveyRules = getNewRules(changes)
+
+      onChange({...updatedStep, choices, surveyRules})
```
hallieswan, line 245: Could `updateChoicesAndRules` be updated to take `step` as a parameter, so that function could be called here?

hallieswan (approved):
Looks good!

PR #771, opened 2023-10-26
Fix font and sizing of the logos with the embedded title

syoung-smallwisdom:
On Chrome the width/height constraints aren’t respected. Instead, make the logos match the original size of the MTB logo.


Erin-Mounts (approved):
looks good

src/assets/logo_open_bridge_large.svg
hallieswan, line 1: This works! However, extra whitespace is added to the right of the logo, which is related to changing the `viewBox` to have a different width than the SVG content, which causes the [SVG content to scale down](https://stackoverflow.com/a/15336348). Instead, we could specify `height` and `width` values that match the values in the original `viewBox`, then the SVG should stay the same size within the container, without adding the extra whitespace to the right of the logo. For example, something like this -- 

```
<svg id="Layer_2" viewBox="0 0 366 55.3" width="55.3" height="366" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
....
```

would be similar to how the `viewBox`, `height`, and `width` were set in the [previous MTB logo](https://github.com/BridgeDigitalHealth/open-bridge-web/blob/62ba868169494993026d3c8bc84f56274f152fdc/src/assets/logo_mtb_large.svg?short_path=350f1f9#L1).

syoung-smallwisdom: Sorry, I'm not understanding. I changed the view port to match the mtb logos which includes the whitespace. Should I be cropping them to fit?

hallieswan (approved):

PR #772, opened 2023-10-27
DHP-1097 Update assessment and survey card to match 2.0 designs

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/DHP-1097

Screenshot 2023-10-27 at 11 08 26 AM

### DIFF
--- a/src/components/assessments/AssessmentCard.tsx
+++ b/src/components/assessments/AssessmentCard.tsx
@@ -25,6 +25,7 @@ const StyledCardContent = styled(CardContent)(({theme}) => ({
   height: theme.spacing(46),
   overflowY: 'scroll',
   paddingBottom: 0,
+  // TODO: syoung 10/27/2023 This is here to enforce showing a scrollbar in Safari b/c by default, Apple hides it.
   '&::-webkit-scrollbar': {
     width: '6px',
   },
--- a/src/components/assessments/AssessmentCard.tsx
+++ b/src/components/assessments/AssessmentCard.tsx
@@ -1,73 +1,38 @@
-import {Box} from '@mui/material'
+import {Card, CardActions} from '@mui/material'
 import CardContent from '@mui/material/CardContent'
 import {styled} from '@mui/material/styles'
 import Typography from '@mui/material/Typography'
-import createStyles from '@mui/styles/createStyles'
-import makeStyles from '@mui/styles/makeStyles'
 import {FunctionComponent} from 'react'
-import {theme} from '../../style/theme'
+import {latoFont, theme} from '../../style/theme'
 import {Assessment} from '../../types/types'
 import AssessmentImage from './AssessmentImage'
 
-const CardContainer = styled(Box, {label: 'cardContainer'})(({theme}) => ({
-  backgroundColor: '#FFFFFF',
+const StyledCard = styled(Card)(({theme}) => ({
+  width: theme.spacing(28),
+  height: theme.spacing(47),
+  borderTop: `5px solid #8FD6FF`,
+  textAlign: 'left',
+  borderRadius: 0,
+  padding: 0,
 
-  boxShadow: '0px 1px 4px rgba(0, 0, 0, 0.15)',
+  display: 'flex',
+  flexDirection: 'column',
+  color: '#353A3F',
+  boxShadow: '0px 4px 4px rgba(0, 0, 0, 0.25)',
 }))
 
-const useStyles = makeStyles(theme =>
-  createStyles({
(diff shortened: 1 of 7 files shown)


src/components/assessments/AssessmentImage.tsx
```diff
@@ -14,12 +15,13 @@ const CardTop = styled(Box, {label: 'StyledCardMedia'})(({theme}) => ({
   flexDirection: 'row',
   '& img': {
     width: '150px',
-    height: '150px',
+    // height: '150px',
```
Erin-Mounts, line 18: remove or TODO

src/components/assessments/AssessmentImage.tsx
```diff
@@ -14,12 +15,13 @@ const CardTop = styled(Box, {label: 'StyledCardMedia'})(({theme}) => ({
...
     backgroundColor: '#f6f6f6', //ALINA TODO PROD CHANGE - remove when images updated to new
```
Erin-Mounts, line 19: Is this comment still relevant?

Erin-Mounts (approved):
Couple of nits/questions, otherwise looks good

src/components/assessments/AssessmentImage.tsx
syoung-smallwisdom: Turns out it is, in that the images are updated to *not* include a background color.

src/components/assessments/AssessmentCard.tsx
```diff
@@ -1,73 +1,38 @@
...
 import {Assessment} from '../../types/types'
 import AssessmentImage from './AssessmentImage'
 
-const CardContainer = styled(Box, {label: 'cardContainer'})(({theme}) => ({
-  backgroundColor: '#FFFFFF',
-
-  boxShadow: '0px 1px 4px rgba(0, 0, 0, 0.15)',
+const StyledCard = styled(Card)(({theme}) => ({
+  width: theme.spacing(28),
+  height: theme.spacing(47),
```
hallieswan, line 12: I did something similar in [this PR](https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/267#discussion_r1218277137), but ultimately decided not to use the theme spacing for width/height, since I wasn't convinced that the component dimensions should change if the spacing was changed in the future, so used fixed values instead. Up to you whether that logic applies here!

syoung-smallwisdom: This is copy/pasted from the static site which was apparently more up-to-date? Really, it would be nice to have these use a shared source for the design of the assessment library but not today. ;)

src/components/assessments/AssessmentCard.tsx
```diff
@@ -1,73 +1,38 @@
...
+const StyledCardContent = styled(CardContent)(({theme}) => ({
+  height: theme.spacing(46),
+  overflowY: 'scroll',
+  paddingBottom: 0,
+  '&::-webkit-scrollbar': {
+    width: '6px',
+  },
+  '&::-webkit-scrollbar-thumb': {
+    backgroundColor: '#C4C4C4',
+  },
```
hallieswan, line 33: Does the scrollbar need to be restyled? If so, we may need to look into a cross-browser solution and consider whether the styling will be reliable. For example, the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar) note the following about `::-webkit-scrollbar` -- 

> Non-standard: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

hallieswan (approved):
Looks good to me!

src/components/assessments/AssessmentDetail.tsx
```diff
@@ -58,7 +58,7 @@ const AssessmentDetail: FunctionComponent<AssessmentDetailProps> = () => {
   let {id} = useParams<{id: string}>()
   const handleError = useErrorHandler()
 
-  const {data, isError, error, isLoading} = useAssessmentWithResources(id)
+  const {data, isError, error, isLoading} = useAssessmentWithResources(id, false)
```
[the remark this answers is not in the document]
syoung-smallwisdom: Correct, it's a required field b/c of the whole different endpoints for local vs. non-local assessments

src/components/assessments/AssessmentCard.tsx
syoung-smallwisdom: Added a TODO comment.

PR #773, opened 2023-10-31
Rename and update privacy policy and terms of service

syoung-smallwisdom:
I decided to go with a generic name for both the privacy policy and terms of service. We will need to replace these with the production versions in both the apps as will as static and studies endpoints.


public/MTB-ToS-v2-210923.pdf
Erin-Mounts, line 1: So, reading the Google doc, I think may be we need to consider whether we should refer to it as "the Open Bridge" vs. just "Open Bridge". "Use the Mobile Toolbox" makes sense in that the name is also a metaphorical description of what it is, whereas "Open Bridge" leans more toward being just a name, and to my mind "Use the Open Bridge" sounds like we're asking them to drive off an open drawbridge.

public/MTBPrivacyPolicy.pdf
Erin-Mounts, line 1: In the Google doc, the date in the change log said "Oct0ber" (with a zero). I've edited it to "October" there but I assume it will need to be re-exported. Sorry, GitHub treats PDFs as straight up binary files and doesn't show line-by-line comparisons where I could suggest an edit in place.

public/privacy_policy.pdf
Erin-Mounts, line 1: I'm guessing this is a copy of MTBPrivacyPolicy.pdf and if so, the same comment would apply.

public/terms_of_service.pdf
Erin-Mounts, line 1: I'm guessing this is a copy of the MTB-ToS-v2... file above and if so, the same comment would apply.

Erin-Mounts (approved):
Couple of comments but otherwise these look good.

PR #776, opened 2023-11-01
update QR code to point at http://install.bridgedigital.health

syoung-smallwisdom:


Erin-Mounts (approved):
looks like a QR code to me

PR #777, opened 2023-11-01
Making branching logic read-only in production

syoung-smallwisdom:
Found issues with the JSON config and how survey rules are being set up. In the interest of moving forward with other concerns, we're making branching logic read-only on production.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -53,7 +53,11 @@ function App() {
         {redirect && <Redirect to={redirect}></Redirect>}
         {/*  <React.StrictMode>*/}
         <FeatureToggleProvider
-          featureToggles={{[features.SURVEY_BUILDER]: true, [features.USERNAME_PASSWORD_LOGIN]: isDevelopment()}}>
+          featureToggles={{
+            [features.SURVEY_BUILDER]: true, 
+            [features.SURVEY_BRANCHING_LOGIC]: isDevelopment(),
+            [features.USERNAME_PASSWORD_LOGIN]: isDevelopment(),
+            }}>
           <Container
             id="outer"
             maxWidth="xl"
--- a/src/components/surveys/survey-branching/SurveyBranching.tsx
+++ b/src/components/surveys/survey-branching/SurveyBranching.tsx
@@ -1,11 +1,12 @@
 import ConfirmationDialog from '@components/widgets/ConfirmationDialog'
-import {Box, styled} from '@mui/material'
+import {Box, styled, Typography} from '@mui/material'
 import {useSurveyAssessment, useSurveyConfig, useUpdateSurveyConfig} from '@services/assessmentHooks'
-import {latoFont} from '@style/theme'
+import {latoFont, theme} from '@style/theme'
 import {ChoiceQuestion, Step, Survey} from '@typedefs/surveys'
 import dagre from 'dagre'
 import React, {FunctionComponent} from 'react'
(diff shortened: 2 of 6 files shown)


src/App.tsx
```diff
@@ -53,7 +53,11 @@ function App() {
         {redirect && <Redirect to={redirect}></Redirect>}
         {/*  <React.StrictMode>*/}
         <FeatureToggleProvider
-          featureToggles={{[features.SURVEY_BUILDER]: true, [features.USERNAME_PASSWORD_LOGIN]: isDevelopment()}}>
+          featureToggles={{
+            [features.SURVEY_BUILDER]: true, 
+            [features.SURVEY_BRANCHING_LOGIC]: isDevelopment(),
+            [features.USERNAME_PASSWORD_LOGIN]: isDevelopment(),
```
Erin-Mounts, line 59: Do we still want to keep this hidden in production? Or does this not do what I think it does?

Erin-Mounts (approved):
One question, otherwise looks good to me

src/App.tsx
syoung-smallwisdom: The features flag allows you to turn features on/off. This just says that it should be "turned off" for production.

How it's actually used is that it is hidden as a "review your branching logic" on the completion page and that if you tab into the branching logic, it's read-only and there is a banner that tells you why its read-only.

syoung-smallwisdom: Actually, not sure what you are asking. I did not change `features.SURVEY_BUILDER` or `features.USERNAME_PASSWORD_LOGIN` - just reformatted for readability.

I added `features.SURVEY_BRANCHING_LOGIC` and set it to only read-write in staging.

src/components/surveys/survey-branching/SurveyBranching.tsx
```diff
@@ -124,9 +125,12 @@ const SurveyBranching: FunctionComponent<SurveyBranchingProps> = () => {
   const [hasObjectChanged, setHasObjectChanged] = React.useState(false)
   const [currentStepIndex, setCurrentStepIndex] = React.useState<number | undefined>(-1)
   const [isHideInput, setIsHideInput] = React.useState(true)
-  const isReadOnly = assessment?.isReadOnly ?? false
   const [nodes, setNodes] = React.useState<Node[]>([])
   const [edges, setEdges] = React.useState<Edge[]>([])
+  const featureToggles = useFeatureToggles<FeatureToggles>()
+  const isBranchingLogicUnlocked = (featureToggles[features.SURVEY_BRANCHING_LOGIC] ?? false)
+  const isAssessmentReadOnly = (assessment?.isReadOnly ?? false)
+  const isReadOnly = isAssessmentReadOnly || !isBranchingLogicUnlocked
```
syoung-smallwisdom, line 133: Make the branching logic read-only if either the assessment is read only or branching logic is locked.

src/App.tsx
Erin-Mounts: I guess I should have been more specific--I was asking if USERNAME_PASSWORD_LOGIN should still be marked as in development or if it's time to turn it on in production, especially now that the "forgot password" link is there.

syoung-smallwisdom: No, that is not production ready. We never resolved the issue of how to store the session token when the users hits refresh.

Erin-Mounts: :+1:

PR #778, opened 2023-11-02
DHP-1038

hallieswan:
https://sagebionetworks.jira.com/browse/DHP-1038

https://github.com/BridgeDigitalHealth/open-bridge-web/assets/26949006/1adf2cc8-3bf0-4539-884d-0bfea66165a9

### DIFF
--- a/src/components/access-settings/AccessSettings.tsx
+++ b/src/components/access-settings/AccessSettings.tsx
@@ -8,7 +8,6 @@ import {Alert, Avatar, Backdrop, Box, Button, CircularProgress, Theme, Typograph
 import makeStyles from '@mui/styles/makeStyles'
 import AccessService from '@services/access.service'
 import ParticipantService from '@services/participants.service'
-import UserService from '@services/user.service'
 import {theme} from '@style/theme'
 import {LoggedInUserClientData, LoggedInUserData, Study} from '@typedefs/types'
 import clsx from 'clsx'
@@ -16,6 +15,7 @@ import React, {FunctionComponent, useRef} from 'react'
 import {useErrorHandler} from 'react-error-boundary'
 import AccessGrid, {Access, NO_ACCESS, getAccessFromRoles, getRolesFromAccess, userHasCoadminAccess} from './AccessGrid'
 import MemberInvite, {NewOrgAccount} from './MemberInvite'
+import {useCreateNewAccount} from './accessHooks'
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
@@ -137,7 +137,6 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
   >()
   const [isAccessLoading, setIsAccessLoading] = React.useState(true)
   const [isUpdating, setIsUpdating] = React.useState(false)
-  const [isCreatingNewMember, setIsCreatingNewMember] = React.useState(false)
   const [isConfirmDeleteOpen, setIsConfirmDeleteOpen] = React.useState(false)
   const [updateError, setUpdateError] = React.useState('')
   const [members, setMembers] = React.useState<LoggedInUserData[]>([])
@@ -154,48 +153,16 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
     [id]
   )
 
-  async function createNewAccount(
-    email: string,
-    firstName: string | undefined,
-    lastName: string | undefined,
-    access: Access,
-    token: string,
-    currentUserOrg: string
-  ) {
-    try {
-      let principalId: string | null = null
-      let synapseFirstName: string | null = null
-      let synapseLastName: string | null = null
-
-      if (AccessService.isSynapseEmail(email)) {
-        const synapseAlias = await AccessService.getAliasFromSynapseByEmail(email)
-        principalId = synapseAlias.principalId
-        synapseFirstName = synapseAlias.firstName
-        synapseLastName = synapseAlias.lastName
-      }
-
-      const demoExternalId = await ParticipantService.signUpForAssessmentDemoStudy(token!)
-
-      const {identifier: userId} = await AccessService.createIndividualAccount(
-        token!,
-        email,
-        principalId,
-        firstName || synapseFirstName,
-        lastName || synapseLastName,
-        currentUserOrg,
-        {demoExternalId},
-        getRolesFromAccess(access)
-      )
-
-      if (!AccessService.isSynapseEmail(email)) {
-        await UserService.sendRequestResetPassword(userId, token!)
-      }
-
-      return [true]
-    } catch (error) {
-      return [false, error]
-    }
-  }
+  const {mutate: inviteUser, isLoading: isCreatingNewMember} = useCreateNewAccount({
+    onSuccess: () => {
+      setNewOrgAccount(CreateNewOrgAccountTemplate())
+      setUpdateToggle(prev => !prev)
+    },
+    onError: error => {
+      const errorString = error.message || error.reason
+      setNewOrgAccount({...newOrgAccount, error: errorString})
+    },
+  })
 
   React.useEffect(() => {
     let isSubscribed = true
@@ -270,31 +237,6 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
 
   const isSelf = (): boolean => {
     return currentMemberAccess?.member.id === id
-  }
-
-  const inviteUser = async (account: NewOrgAccount) => {
-    setIsCreatingNewMember(true)
-    if (!account.email) {
-      setNewOrgAccount({...account, error: 'No email provided'})
-      setIsCreatingNewMember(false)
-      return
-    }
-    const [success, error] = await createNewAccount(
-      account.email,
-      account.firstName,
-      account.lastName,
-      account.access,
-      token!,
-      orgMembership!
-    )
-    if (success) {
-      setNewOrgAccount({...account, isAdded: true})
-      setUpdateToggle(prev => !prev)
-    } else {
-      const errorString = error.message || error.reason
-      setNewOrgAccount({...account, error: errorString})
-    }
-    setIsCreatingNewMember(false)
   }
 
   return (
@@ -458,7 +400,7 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
               onClick={() => {
                 scollToRef.current?.scrollIntoView()
                 setCurrentMemberAccess(undefined)
-                inviteUser(newOrgAccount)
+                inviteUser({account: newOrgAccount, currentUserOrg: orgMembership!})
               }}
               color="primary"
               variant="contained">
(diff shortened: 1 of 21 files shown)


src/components/access-settings/AccessSettings.tsx
```diff
@@ -57,13 +58,8 @@ const useStyles = makeStyles((theme: Theme) => ({
...
 
-function getNameDisplay({
-  firstName,
-  lastName,
-
-  synapseUserId,
-}: LoggedInUserData): string {
-  const name = firstName || lastName ? [firstName, lastName].join(' ') : synapseUserId
+function getNameDisplay({firstName, lastName, email, synapseUserId}: LoggedInUserData): string {
+  const name = firstName || lastName ? [firstName, lastName].join(' ') : synapseUserId || email
```
hallieswan, line 62: Fallback to email address when member doesn't have a first/last name or a synapseUserId

src/components/access-settings/AccessSettings.tsx
```diff
@@ -157,23 +154,43 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
...
+      let principalId: string | null = null
+      let synapseFirstName: string | null = null
+      let synapseLastName: string | null = null
+
+      if (AccessService.isSynapseEmail(email)) {
+        const synapseAlias = await AccessService.getAliasFromSynapseByEmail(email)
+        principalId = synapseAlias.principalId
+        synapseFirstName = synapseAlias.firstName
+        synapseLastName = synapseAlias.lastName
+      }
```
hallieswan, line 175: Continue getting Synapse profile info when a synapse email address is provided

src/components/access-settings/AccessSettings.tsx
```diff
@@ -157,23 +154,43 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
...
+        firstName || synapseFirstName,
+        lastName || synapseLastName,
         currentUserOrg,
         {demoExternalId},
         getRolesFromAccess(access)
       )
 
+      if (!AccessService.isSynapseEmail(email)) {
+        await UserService.sendRequestResetPassword(userId, token!)
+      }
```
hallieswan, line 192: Only send a request reset password email to an account created with a non-Synapse email address

src/components/access-settings/AccessSettings.tsx
```diff
@@ -320,7 +347,7 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
               onClick={() => setIsAddingNewMember(true)}
               variant="contained"
               color="primary"
-              disabled={isAddingNewMember}>
+              disabled={isAddingNewMember || isAccessLoading}>
```
hallieswan, line 350: Ensure that the user does not click the "Add New Member" button before the members response returns -- otherwise, the "Add New Member" panel is briefly shown then removed by the `updateAccess` call which removes the "Add New Member" panel when the members are updated

src/components/access-settings/AccessSettings.tsx
```diff
@@ -434,8 +462,11 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
               }}
               color="primary"
               variant="contained">
-              Save Changes
+              {isCreatingNewMember ? 'Saving...' : 'Save Changes'}
             </Button>
+            <Backdrop open={isCreatingNewMember} sx={{zIndex: theme => theme.zIndex.drawer + 1}}>
+              <CircularProgress />
+            </Backdrop>
```
hallieswan, line 469: Show a loader while the new member is created

src/components/surveys/survey-design/question-edit/phone-subcontrols/Completion.test.tsx
```diff
@@ -25,6 +25,11 @@ const step: Step = {
   },
 }
 
+afterEach(() => {
+  jest.runOnlyPendingTimers()
+  jest.useRealTimers()
+})
```
hallieswan, line 31: `renderSurveyQuestionComponent`, which is used to render the survey components, calls `jest.useFakeTimers()`. Real timers should be restored after each test

src/components/surveys/survey-design/question-edit/rhs-subcontrols/Year.test.tsx
```diff
@@ -46,6 +46,7 @@ function setUp(step: YearQuestion) {
 
 const setUpWithChangeMock = (step: YearQuestion) => {
   const onChangeMock = jest.fn()
+  jest.useFakeTimers()
```
hallieswan, line 49: Update this set up method to explicitly use fake timers, rather than depending on a previous test to set fake timers.

src/services/access.service.ts
```diff
@@ -9,11 +9,22 @@ const AccessService = {
...
   getDetailedAccountsForOrg,
   getIndividualAccount,
+  isSynapseEmail,
   updateIndividualAccountRoles,
 }
 
-async function getAliasFromSynapseByEmail(synapseEmailAddress: string): Promise<any> {
-  if (synapseEmailAddress.indexOf('@synapse.org') === -1 && synapseEmailAddress.indexOf('@synapse.org') > -1) {
+function isSynapseEmail(emailAddress: string): boolean {
+  return emailAddress.replace(new RegExp('.*@'), '') === 'synapse.org'
```
hallieswan, line 17: Previous expression always evaluated to false -- update expression to distinguish Synapse vs non-Synapse email addresses

src/services/access.service.ts
```diff
@@ -95,9 +111,8 @@ async function createIndividualAccount(
     clientData,
   }
   const endpoint = constants.endpoints.bridgeAccount.replace(':id', '')
-  const result = await Utility.callEndpoint<any>(endpoint, 'POST', postData, token)
-
-  return result.ok
+  const result = await Utility.callEndpoint<IdentifierHolder>(endpoint, 'POST', postData, token)
+  return result.data
```
hallieswan, line 115: Return the newly created user's identifier, so that the request reset password email can be sent to them

src/components/access-settings/AccessSettings.tsx
nickgros: I see, this is a new account, so there is no password, and the user cannot log in with Synapse OIDC. Thus it must be reset.

This kinda feels like logic that shouldn't be in the client, but I don't think it's necessarily a big deal--this is probably more 'agile' since this service already exists

src/components/access-settings/AccessSettings.tsx
```diff
@@ -256,18 +273,28 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
...
     }
-    const [success, error] = await createNewAccount(account.email, account.access, token!, orgMembership!)
+    const [success, error] = await createNewAccount(
+      account.email,
+      account.firstName,
+      account.lastName,
+      account.access,
+      token!,
+      orgMembership!
+    )
```
nickgros, line 289: I feel like it's a weird pattern to swallow the error in `createNewAccount` but still return it here. I was expecting you would need `try ... finally { setIsCreatingNewMember(false) }` but I see that `createNewAccount` can never throw, so this is logically fine. Just a little weird.

If you have react-query in this project, I would suggest capturing `createNewAccount` in a custom hook that wraps `useMutation`, then that hook will just provide state to replace `isCreatingNewMember` and the error.

src/components/access-settings/AccessSettings.tsx
nickgros: This component is kind of a useEffect nightmare but I would have to know more to understand if/how it could be improved to make any suggestions.

src/components/access-settings/AccessSettings.tsx
nickgros: nice use of the theme to calculate an appropriate zIndex

src/services/access.service.ts
nickgros: Is there a reason you can't use `emailAddress.endsWith('@synapse.org')`? I think a RegEx is overkill here since you're not really doing any validation

src/services/access.service.ts
```diff
@@ -9,11 +9,22 @@ const AccessService = {
...
+export type SynapseAlias = {
+  principalId: string
+  firstName: string
+  lastName: string
+}
+
+async function getAliasFromSynapseByEmail(synapseEmailAddress: string): Promise<SynapseAlias> {
+  if (!isSynapseEmail(synapseEmailAddress)) {
     return Promise.reject({message: 'the email should be ...@synapse.org'})
   }
```
nickgros, line 29: This doesn't need to be asynchronous, suggest throwing an error instead of rejecting a promise.

syoung-smallwisdom (approved):
looks good

src/components/access-settings/AccessSettings.tsx
hallieswan: Agreed! I opened [a ticket](https://sagebionetworks.jira.com/browse/DHP-1145) with some ideas for simplifying this component.

src/components/access-settings/AccessSettings.tsx
hallieswan: Agreed, not ideal -- we decided on this solution as a short-term stopgap to give study admins a self-service method to set up accounts for users with non-Synapse email addresses, until a more robust solution can be designed/implemented

src/services/access.service.ts
hallieswan: Good call! I'll update in the next commit

src/components/access-settings/AccessSettings.tsx
hallieswan: Good idea! Done in the next commit

src/components/access-settings/AccessSettings.tsx
```diff
@@ -154,48 +153,16 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
     [id]
   )
 
-  async function createNewAccount(
```
hallieswan: Moved into custom hook `useCreateNewAccount`

src/components/access-settings/AccessSettings.tsx
```diff
@@ -154,48 +153,16 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
...
-        await UserService.sendRequestResetPassword(userId, token!)
-      }
-
-      return [true]
-    } catch (error) {
-      return [false, error]
-    }
-  }
+  const {mutate: inviteUser, isLoading: isCreatingNewMember} = useCreateNewAccount({
+    onSuccess: () => {
```
hallieswan, line 157: `onSuccess` and `onError` replicate the behavior in the old `inviteUser` method, except the `newOrgAccount` is reset to the default template after an account is successfully created, which prevents the last added user's info from pre-populating fields when the add new member form is next opened

src/components/access-settings/AccessSettings.tsx
```diff
@@ -272,31 +239,6 @@ const AccessSettings: FunctionComponent<{study: Study}> = ({study}) => {
     return currentMemberAccess?.member.id === id
   }
 
-  const inviteUser = async (account: NewOrgAccount) => {
-    setIsCreatingNewMember(true)
-    if (!account.email) {
```
hallieswan, line 259: Moved into createNewAccount, so that all errors are captured in the hook

nickgros (approved):
Looks great!

PR #783, opened 2024-01-03
Use the standard header for sign-in for both ARC and OpenBridge

syoung-smallwisdom:
I don't know that there is a JIRA ticket for this, but it's to solve a deep-linking issue where the re-direct to the sign-in page was only set up for MTB and never for ARC.

Deep linking when *not* signed in will hide the page and show the "LOG IN" button - this now works the same way for both ARC dashboard and Open Bridge.


Erin-Mounts (commented):
Looks good to me.

It looks like the code must have been added deliberately to make this distinction and treat ARC differently, though. Do you know why that would have been done?

syoung-smallwisdom:
> Looks good to me.
> 
> It looks like the code must have been added deliberately to make this distinction and treat ARC differently, though. Do you know why that would have been done?

As far as I can tell, it's branding. Or more specifically, because there was a fairly significant redesign and ARC didn't want anything to change. We can hold off on this until I can confirm with Ann that it's something she wants us to do.

PR #784, opened 2024-01-05
Refactor the logos and title branding to allow for supporting multiple websites

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/DHP-1011

This still keeps the 1:1 branding of appId to website, but allows for using the app-id to set the logos for the site without long lists of import statements.

This should also allow us to publish both arc and open-bridge using the same workflow because it fixes some issues with the tab icon in the manifest by setting that dynamically.

Note: I still need an SVG logo for WashU with appropriate sizing in order to update the logo in the browser tab bar and the header.


src/types/constants.ts
```diff
@@ -50,6 +50,27 @@ const constants = {
...
+      iconPrefix: '',
+    },
+    {
+      appId: ARC_APP_ID,
+      title: 'ARC',
+      logo: 'arc',
+      iconPrefix: 'arc_',
+    },
+    {
+      appId: INV_ARC_ID,
```
Erin-Mounts, line 67: Should this be INV_ARC_APP_ID for consistency?

Erin-Mounts (approved):
Couple of nits, otherwise looks good

src/types/constants.ts
syoung-smallwisdom: Probably. Renaming is kinda kludgy in React though so I'd prefer not to blanket rename it. I can add a TODO.

